PR #3, opened 2021-06-23
Enable AutoCompleteConfig feature

tanmayv25:
Retains all the autofiller features.. However, it does not auto-detect plan file. The triton core requires the file to be named "model.plan" for AutoCompleteConfig to construct the full configuration. We can fix this with the move to clean-ups for autofiller in triton core. Added a comment for the same.

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -223,6 +223,28 @@ class ModelState : public TensorRTModel {
 
   // Auto-complete the model configuration
   TRITONSERVER_Error* AutoCompleteConfig();
+  TRITONSERVER_Error* AutoCompleteConfigHelper(const std::string& model_path);
+  TRITONSERVER_Error* GetMaxSupportedBatchSize(
+      nvinfer1::ICudaEngine* engine, const int num_profiles,
+      int* max_batch_size);
+  TRITONSERVER_Error* GetProfileIndices(
+      const int num_profiles, std::set<int>* profile_indices);
+  TRITONSERVER_Error* GetProfileMaxBatchSize(
+      nvinfer1::ICudaEngine* engine, int profile_index, int* max_batch_size);
+  TRITONSERVER_Error* ExtractBatchHintFromIOConfig(
+      nvinfer1::ICudaEngine* engine, const std::string& tensor_name,
+      const common::TritonJson::Value& dims, bool* config_batch_hint);
+  TRITONSERVER_Error* GetRefIO(
+      const bool is_input, nvinfer1::ICudaEngine* engine,
+      triton::common::TritonJson::Value* ref_io);
+  TRITONSERVER_Error* InitIODims(
+      nvinfer1::ICudaEngine* engine, nvinfer1::Dims& dims,
+      bool is_shape_binding, triton::common::TritonJson::Value* io);
+  TRITONSERVER_Error* FixIO(
+      nvinfer1::ICudaEngine* engine,
+      triton::common::TritonJson::Value& reference_ios,
+      triton::common::TritonJson::Value* mutable_ios);
+
 
   // Validate that model configuration is supported by this backend.
   TRITONSERVER_Error* ValidateModelConfig();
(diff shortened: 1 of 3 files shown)


deadeyegoodwin (approved):

CoderHam:
@tanmayv25 I recently encountered a bug in the autocomplete when using a TRT model (legacy backend API). When I specify the count in instance group in the model config.pbtxt file and load the model with --strict-model-config=false it adds an additional instance group field and hence loads an additional instance of the model. 
Can you confirm moving to the new Backend API solves this problem?

src/tensorrt.cc
```diff
@@ -394,17 +416,502 @@ ModelState::ParseParameters()
...
+    }
+
+    triton::common::TritonJson::Value outputs;
+    if (ModelConfig().Find("output", &outputs)) {
+      output_cnt = outputs.ArraySize();
+    }
+  }
+
+  // For batching support, the number of dimensions specified in model config
+  // match should be 1 less than the number of dimensions specified in engine.
```
CoderHam, line 489: Appears to be a typo here

src/tensorrt.cc
```diff
@@ -394,17 +416,502 @@ ModelState::ParseParameters()
...
+    // proven otherwise.
+    RETURN_IF_ERROR(
+        GetMaxSupportedBatchSize(engine, num_profiles, &max_batch_size));
+  }
+
+  if (config_batch_hint && max_batch_size == 0) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        (std::string("unable to autofill for '") + Name() +
+         "', model tensor  shape configuration hints for dynamic batching "
```
CoderHam, line 550: extra space here

CoderHam: For consistency lets stick to a template for the autofill error: https://github.com/triton-inference-server/onnxruntime_backend/blob/main/src/onnxruntime.cc#L536-L539

src/tensorrt.cc
```diff
@@ -394,17 +416,502 @@ ModelState::ParseParameters()
...
+         "dims or none.")
+            .c_str());
+  }
+
+  int max_batch_size = 0;
+  if (engine->hasImplicitBatchDimension()) {
+    // If engine has implicit batch dimension then retrieve the value and exit
+    max_batch_size = engine->getMaxBatchSize();
+  } else {
+    // Assuming the first dimension to be batch dimension, until and unless
```
CoderHam, line 540: "until and unless proven otherwise" should be removed since we do not support the case where batch dims is not the first dims.

CoderHam (changes requested):

src/tensorrt.cc
tanmayv25: For the non-batching case the first dim is not a batch dimension. The autofiller does not know whether or not the plan file supports batching. I will make the language clearer.

GuanLuo (approved):

tanmayv25:
@CoderHam  There is a sing instance_group when my confiig.pbtxt is :
```

instance_group {
count: 2
}
```

Generated config:

```
{
    "name": "plan_float32_float32_float32",
    "platform": "tensorrt_plan",
    "backend": "tensorrt",
    "version_policy": {
        "latest": {
            "num_versions": 1
        }
    },
    "max_batch_size": 8,
    "input": [
        {
            "name": "INPUT0",
            "data_type": "TYPE_FP32",
            "dims": [
                16
            ],
            "is_shape_tensor": false
        },
        {
            "name": "INPUT1",
            "data_type": "TYPE_FP32",
            "dims": [
                16
            ],
            "is_shape_tensor": false
        }
    ],
    "output": [
        {
            "name": "OUTPUT0",
            "data_type": "TYPE_FP32",
            "dims": [
                16
            ],
            "is_shape_tensor": false
        },
        {
            "name": "OUTPUT1",
            "data_type": "TYPE_FP32",
            "dims": [
                16
            ],
            "is_shape_tensor": false
        }
    ],
    "batch_input": [],
    "batch_output": [],
    "optimization": {
        "priority": "PRIORITY_DEFAULT",
        "input_pinned_memory": {
            "enable": true
        },
        "output_pinned_memory": {
            "enable": true
        },
        "gather_kernel_buffer_threshold": 0,
        "eager_batching": false
    },
    "instance_group": [
        {
            "name": "plan_float32_float32_float32_0",
            "kind": "KIND_GPU",
            "count": 2,
            "gpus": [
                0
            ],
            "secondary_devices": [],
            "profile": [],
            "passive": false,
            "host_policy": ""
        }
    ],
    "default_model_filename": "model.plan",
    "cc_model_filenames": {},
    "metric_tags": {},
    "parameters": {},
    "model_warmup": []
}
```

CoderHam (approved):

CoderHam:
Great, I was hoping that moving to the new Backend API will fix the issue. Thanks!

PR #5, opened 2021-08-05
Add support for DLA and include zero-copy fix

tanmayv25:
Includes other minor changes to bring the backend at parity with current backend.


CoderHam (approved):

CoderHam:
@tanmayv25 as soon as we move to JP4.6, we will be able to add / enable the DLA test. I can share a build with you to run the same manually

PR #11, opened 2021-09-03
Fix the extra context index in the queue

tanmayv25:
This prevents the HtoD overwrites which causes inaccurate values.

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -2906,6 +2906,7 @@ ModelInstanceState::RegisterContexts()
       context_queue_.Put(context_idx++);
     }
   }
+  next_context_idx_ = context_queue_.Get();
 }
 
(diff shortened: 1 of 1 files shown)


GuanLuo (approved):
How could the TRT backend before this change work fine? Seems like uninitialized variable was in used before this change

tanmayv25:
The context index is just a placeholder index here whose value is not important. I used the context queue just to maintain some parallels with TRT core backend. The entire queue can be replaced with a semaphore that works as a gating. 
Once the backend is stabilized, I do plan to revisit it for some cleanups.

Before there were just two conflicting executions of the model because the queue would have a random valued index and a zero index.

GuanLuo:
I see, it is no longer being used to select the context (instance) to run the requests as the instance has been selected up front

PR #14, opened 2021-10-06
Initialize BackendInputCollector with coalesce input argument

GuanLuo:
TRT backend is an example of explicitly enable request input coalescent if requested via backend config. TRT backend is the only backend that require this optimization and will only apply the same change to other backends if requested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,10 @@ backend that are then applied to all models that use the backend.
 
 ##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>
 
-Instruct TensorRT to consider the request inputs as one contiguous buffer if
-their memory addresses align with each other.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,6 +45,16 @@ available in the main [server](https://github.com/triton-inference-server/server
...
 
+## Command-line Options
+
+The command-line options configure properties of the TensorRT
+backend that are then applied to all models that use the backend.
+
+##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>
+
+Instruct TensorRT to consider the request inputs as one contiguous buffer if
+their memory addresses align with each other.
```
deadeyegoodwin, line 56: Mention that the default is "false". Also say something like "This option should only be enabled if all requests' input tensors are allocated from the same memory region."

PR #16, opened 2021-10-15
WIP: Allows greater overlap between multiple instance runs

tanmayv25:
This fix is a fallout from https://github.com/triton-inference-server/tensorrt_backend/commit/5e3df3daf566fc25c5dfbc7e9f0e7adc7249c407.

Earlier the issue thread would wait for current instance to consume its input before jumping to next. With this fix the issue thread jumps to the next instance just after enqueueing request.

The below model issues inputConsumed at the  end of its execution and hence is easy to visualize. The instance count is 2.

### Before Fix
2 Instances: Before Fix

### After Fix
2 Instances: After Fix

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -1206,7 +1206,6 @@ class ModelInstanceState : public TensorRTModelInstance {
   std::thread completion_thread_;
 
   triton::common::SyncQueue<size_t> context_queue_;
-  size_t next_context_idx_;
 
   // The details needed by the completion thread to finalize the
   // response for a model execution.
@@ -1559,7 +1558,7 @@ ModelInstanceState::ProcessRequests(
        std::to_string(request_count) + " requests")
           .c_str());
 
-  auto context_idx = next_context_idx_;
+  auto context_idx = context_queue_.Get();
 
   Run(requests, request_count, context_idx);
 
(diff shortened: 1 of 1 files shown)


src/tensorrt.cc
```diff
@@ -1559,7 +1558,7 @@ ModelInstanceState::ProcessRequests(
        std::to_string(request_count) + " requests")
           .c_str());
 
-  auto context_idx = next_context_idx_;
+  auto context_idx = context_queue_.Get();
```
GuanLuo, line 1561: So the issue thread will always run ahead and get the next batch even if there may not be available context? Sounds like the timing for queuing will be count as compute

GuanLuo: Yes, I think that is the main concern that we have, the same concern was raised when we add eager batching support. There are tradeoffs and I don't know which is preferred.

GuanLuo: Actually, wouldn't this change behave the same as if eager batching is enabled?

tanmayv25: Yes it is similar.. Let me do some thinking to get to a better solution..

geraldstanje:
hi @tanmayv25 is image you posted above from nvidia nsight? could you please share some instructions to how someone can run nvidia nsight with triton inference server container?
image

PR #17, opened 2021-10-19
Fix handling of ragged batch inputs for zero copy

tanmayv25:


CoderHam (changes requested):

src/tensorrt.cc
```diff
@@ -3415,13 +3416,20 @@ ModelInstanceState::InitializeBatchInputBindings(
...
           io_binding_infos_[next_buffer_binding_set_][io_index];
 
-      // FIXME.. Don't allocate new backend memory when the original
-      // memory is writable. zero-copy case
       BackendMemory* bm;
-      BackendMemory::Create(
-          model_state_->TritonMemoryManager(),
-          {BackendMemory::AllocationType::CPU_PINNED_POOL},
-          0 /* memory_type_id */, io_binding_info.byte_size_, &bm);
+      if (io_binding_info.memory_type_ != TRITONSERVER_MEMORY_GPU) {
```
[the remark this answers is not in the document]
CoderHam: Agreed. We definitely need more exhaustive testing for jetson.

src/tensorrt.cc
```diff
@@ -3415,13 +3416,20 @@ ModelInstanceState::InitializeBatchInputBindings(
...
+            BackendMemory::AllocationType::CPU_PINNED_POOL,
+            0 /* memory_type_id */, io_binding_info.buffer_,
+            io_binding_info.byte_size_, &bm);
+      } else {
+        BackendMemory::Create(
+            model_state_->TritonMemoryManager(),
+            {BackendMemory::AllocationType::CPU_PINNED_POOL},
+            0 /* memory_type_id */, io_binding_info.byte_size_, &bm);
+      }
       io_binding_info.batch_input_.reset(new BatchInputData(batch_input, bm));
```
GuanLuo, line 3433: Why this fixes the issue? Does the execute function below do something special on copying batch input to binding buffer when it sees it is zero copy? i.e. "writing to CPU batch input buffer -> copy to binding buffer" becomes "writing to CPU batch input buffer" as the CPU buffer to be written is already the binding buffer.

tanmayv25: >  i.e. "writing to CPU batch input buffer -> copy to binding buffer" becomes "writing to CPU batch input buffer" as the CPU buffer to be written is already the binding buffer.

So, the answer to above question is a yes.

PR #19, opened 2021-10-29
Add implicit state support

Tabrizian:
Adds implicit state support for tensorrt backend.

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -220,20 +220,9 @@ class ModelState : public TensorRTModel {
 
   void DisableEngineSharing() { engine_sharing_ = false; }
   bool IsEngineSharingEnabled() { return engine_sharing_; }
-  const std::map<std::string, std::pair<int64_t, int64_t>>& ModelOutputs()
-  {
-    return model_outputs_;
-  }
 
  private:
   ModelState(TRITONBACKEND_Model* triton_model);
-
-  // model_outputs is a map that contains unique outputs that the model must
-  // provide. In the model configuration, the output in the state configuration
-  // can have intersection with the outputs section of the model. If an output
-  // is specified both in the output section and state section, it indicates
-  // that the backend must return the output state to the client too.
-  std::map<std::string, std::pair<int64_t, int64_t>> model_outputs_;
 
   // Auto-complete the model configuration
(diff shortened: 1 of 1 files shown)


src/tensorrt.cc
```diff
@@ -220,10 +220,21 @@ class ModelState : public TensorRTModel {
...
+  }
 
  private:
   ModelState(TRITONBACKEND_Model* triton_model);
 
+  // model_outputs is a map that contains unique outputs that the model must
+  // provide. In the model configuration, the output in the state configuration
+  // can have intersection with the outputs section of the model. If an output
+  // is specified both in the output section and state section, it indicates
+  // that the backend must return the output state to the client too.
```
tanmayv25, line 235: I don't see any value in having these in64_t flags. These are just for detecting whether the unique outputs are state output and requested output.

How about adding these properties in `IOBindingInfo`? `is_state_tensor` and `is_requested_output_tensor`?

When parsing the buffers while placing outputs in response, we can utilize this information to call `ProcessTensor` or `ProcessStateTensor` acordingly.

tanmayv25 (changes requested):

src/tensorrt.cc
```diff
@@ -3373,6 +3450,41 @@ ModelInstanceState::InitializeSequenceControlInputBindings(
...
+        RETURN_IF_ERROR(io.MemberAsString("data_type", &io_datatype));
+
+        common::TritonJson::Value model_config_dims;
+        RETURN_IF_ERROR(io.MemberAsArray("dims", &model_config_dims));
+
+        RETURN_IF_ERROR(InitializeExecuteInputBinding(
+            input_name, io_datatype, model_config_dims, false /* is_control */,
+            false /* is_ragged */));
+
+        RETURN_IF_ERROR(InitializeExecuteOutputBinding(
```
tanmayv25, line 3479: Also, isn't the logic calling InitializeExecuteOutputBinding twice for outputs that are in state section?

src/tensorrt.cc
Tabrizian: Sounds good, fixed.

src/tensorrt.cc
Tabrizian: Good catch, fixed.

GuanLuo (approved):

PR #20, opened 2021-11-05
Fix the binding initializing logic to correctly detect repeated tensors

tanmayv25:

PR #21, opened 2021-11-06
Fix the multi-instance overlap in TRT backend

tanmayv25:
This PR extends PR #16.

Please review only latest commit. 
cb020088fafe6e250c5faa563be524ddcae0b773

The first commit is part of the PR #20.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,7 @@ add_library(
   src/tensorrt_model_instance.h
   src/tensorrt_utils.cc
   src/tensorrt_utils.h
+  src/semaphore.h
   src/loader.cc
   src/loader.h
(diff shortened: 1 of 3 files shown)


tanmayv25:
### Before

4_instance_before

### After

4_instance_after

Four instances are properly being executed.

GuanLuo (approved):
The change looks good, I think round-robin over the `semaphore_list_` in line 1643 is okay, I don't think it is possible that the context will be completed out of order.

tanmayv25:
> The change looks good, I think round-robin over the `semaphore_list_` in line 1643 is okay, I don't think it is possible that the context will be completed out of order.

Exactly! The ops are enqueued to the same stream!

PR #25, opened 2022-03-10
Load Shared Libraries Dynamically

dyastremsky:
Allow user to load shared TensorRT libraries via the backend configuration.

Related server ticket: https://github.com/triton-inference-server/server/pull/4039


src/tensorrt.cc
```diff
@@ -46,6 +46,14 @@
 #include <thread>
 #include <unordered_map>
 
+#ifdef _MSC_VER
+#include <windows.h>
+// undefine the max macro in windows.h, so std::max can compile
+#undef max
```
tanmayv25, line 52: For consistency, define NOMINMAX before including windows.h . We do this everywhere else.
See here for example: https://github.com/triton-inference-server/server/blob/main/src/main.cc#L28

src/tensorrt.cc
```diff
@@ -5275,6 +5283,31 @@ ModelInstanceState::SetCudaGraphShape(
...
+#else
+  void* handle = LoadLibrary(path.c_str());
+#endif
+#else
+  void* handle = dlopen(path.c_str(), RTLD_LAZY);
+#endif
+
+  if (handle == nullptr) {
+#ifdef _MSC_VER
+    std::cout << "Could not load plugin library: " << path << std::endl;
```
tanmayv25, line 5301: Is there a way to figure out error details of LoadLibraryA / LoadLibrary?

tanmayv25, line 5301: std::cerr ??

src/tensorrt.cc
```diff
@@ -5275,6 +5283,31 @@ ModelInstanceState::SetCudaGraphShape(
 
 extern "C" {
 
+inline void
```
tanmayv25, line 5286: Instead of setting return type void, should we return TRITONSERVER_Error* and let the caller decide how to handle the error ?

src/tensorrt.cc
```diff
@@ -5359,12 +5377,44 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
...
+  bool success = true;
+  std::once_flag onceFlag;
+  {
+    std::call_once(onceFlag, [&success] {
+      LOG_MESSAGE(TRITONSERVER_LOG_VERBOSE, "Registering TensorRT Plugins");
+      success = initLibNvInferPlugins(&tensorrt_logger, "");
+    });
+  }
+  if (!success) {
+    TRITONSERVER_ErrorNew(
```
tanmayv25, line 5414: We can log an error message when failing to register the plugin.

src/tensorrt.cc
```diff
@@ -5359,12 +5377,44 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
...
+
+    if (cmdline.Find("plugins", &value)) {
+      RETURN_IF_ERROR(value.AsString(&value_str));
+      size_t pos = 0;
+      std::string plugin;
+      // Load individual plugins
+      while (value_str.length() > 0) {
+        pos = value_str.find(";");
+        plugin = value_str.substr(0, pos);
+        LoadPlugin(plugin);
```
tanmayv25, line 5395: Take the error object from LoadPlugin and log the error message here(using LOG_MESSAGE is better). I think we can still continue without aborting.

src/tensorrt.cc
```diff
@@ -5359,12 +5377,44 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
...
     triton::common::TritonJson::Value value;
     std::string value_str;
+
     if (cmdline.Find("coalesce-request-input", &value)) {
       RETURN_IF_ERROR(value.AsString(&value_str));
       RETURN_IF_ERROR(
           ParseBoolValue(value_str, &lconfig->coalesce_request_input_));
     }
+
+    if (cmdline.Find("plugins", &value)) {
```
tanmayv25, line 5387: Shouldn't this be also a part of once_flag ? If there are multiple TensorRT models hosted in the server then it is not necessary to load the plugins with each load.

tanmayv25 (changes requested):

src/tensorrt.cc
dyastremsky: Good question. Technically, yes. However, it can get messy. We'd likely want to use GetLastError, with the documentation here: https://docs.microsoft.com/en-us/windows/win32/debug/retrieving-the-last-error-code

A busy StackOverflow post with different solutions (and critiques of those solutions e.g. not supporting Unicode or Ansi/MBSC, creating memory leaks, potentially getting the error too late, etc.): https://stackoverflow.com/a/45565001. The link goes to what seems to be a good approach, though I think it'd require some research to be sure. We might still run into edge cases. Let me know if you think the benefit-risk tradeoff is worth it; if so, I'll implement it.

src/tensorrt.cc
dyastremsky: Is TRITONBACKEND_Initialize only called once? That seems to be what the backend header file implies, but I don't know if there's an exception. If so, could we get rid of once_flag in general?

If not, I can move the whole backend-config reading into the once_flag (probably in its own function).

src/tensorrt.cc
```diff
@@ -5275,6 +5282,37 @@ ModelInstanceState::SetCudaGraphShape(
...
+        (std::string("Could not load plugin library: ") + path).c_str());
+#else
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_NOT_FOUND,
+        (std::string("Could not load plugin library: ") + path +
+         ", due to: " + dlerror())
+            .c_str());
+#endif
+  } else {
+    std::cout << "Successfully loaded plugin library: " << path << std::endl;
```
Tabrizian, line 5311: I think we should be using the `LOG_MESSAGE` instead of `std::cout`.

Tabrizian (changes requested):

src/tensorrt.cc
dyastremsky: Great suggestion! Fixed.

src/tensorrt.cc
dyastremsky: Updated. :)

src/tensorrt.cc
dyastremsky: Good point! Switched it to logging as per your other feedback.

src/tensorrt.cc
dyastremsky: Done, thanks. Also updated to use the same ifdef flag for consistency (#ifdef _WIN32).

src/tensorrt.cc
dyastremsky: I confirmed that even with multiple models, the plugins are only loaded once.

It looks like this whole function only gets called once, so removing the once_flag seems like a good idea. Let me know if there's an edge or breaking case you can think of, otherwise I'll get delete it.

Tabrizian (approved):

src/tensorrt.cc
tanmayv25: Let's leave it as such for now. If it was simple we could have returned a more specific error but it is fine as it is now.

src/tensorrt.cc
tanmayv25: >Is TRITONBACKEND_Initialize only called once? 

That's correct.

> I confirmed that even with multiple models, the plugins are only loaded once.

I was thinking something else.

> It looks like this whole function only gets called once, so removing the once_flag seems like a good idea. Let me know if there's an edge or breaking case you can think of, otherwise I'll delete it.

One edge case is that what if we have another backend (custom backend) that uses TRT plugins. So, for now let's leave the register part in the once_flag and load these plugins once. In summary let's leave them as such.

tanmayv25 (approved):

src/tensorrt.cc
```diff
@@ -5275,6 +5282,39 @@ ModelInstanceState::SetCudaGraphShape(
...
+  if (handle == nullptr) {
+#ifdef _WIN32
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_NOT_FOUND,
+        (std::string("Could not load plugin library: ") + path).c_str());
+#else
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_NOT_FOUND,
+        (std::string("Could not load plugin library: ") + path +
+         ", due to: " + dlerror())
```
CoderHam, line 5307: Why not keep the error messages symmetric?

src/tensorrt.cc
```diff
@@ -5359,12 +5384,48 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
...
           ParseBoolValue(value_str, &lconfig->coalesce_request_input_));
     }
+
+    if (cmdline.Find("plugins", &value)) {
+      RETURN_IF_ERROR(value.AsString(&value_str));
+      size_t pos = 0;
+      std::string plugin;
+      // Load individual plugins
+      while (value_str.length() > 0) {
+        pos = value_str.find(";");
```
CoderHam, line 5400: What if there is only 1 plugin and no ';'?

CoderHam (changes requested):

src/tensorrt.cc
dyastremsky: std::string::npos gets returned. substr() will then return all characters until the end. It passes tests for both one and two plugins.

src/tensorrt.cc
dyastremsky: Please see the conversation here: https://github.com/triton-inference-server/tensorrt_backend/pull/25#discussion_r835546722

Essentially, pulling the error for LoadLibrary/LoadLibraryA gets messy. We can do it, but the code would be more complicated and possibly introduce edge cases in terms of leaks and functionality. dlerror() is easy to call, so I think we should provide more information though since we can.

CoderHam (approved):

src/tensorrt.cc
```diff
@@ -5275,6 +5282,39 @@ ModelInstanceState::SetCudaGraphShape(
...
+LoadPlugin(const std::string& path)
+{
+#ifdef _WIN32
+#ifdef UNICODE
+  void* handle = LoadLibraryA(path.c_str());
+#else
+  void* handle = LoadLibrary(path.c_str());
+#endif
+#else
+  void* handle = dlopen(path.c_str(), RTLD_LAZY);
```
GuanLuo, line 5295: Shouldn't the `handle` be stored and release when the backend is being finalized?

dyastremsky: Good point, thank you. We'll probably need to store a list of handles.

Since the core library has been moved into its own repo, we can also switch to using it here and get rid of the duplicated shared_library.h file in this and other repos. I'll switch to that approach.

GuanLuo: I agree that we may expose the `shared_library.h` util for all other repos (and move it to `common` repo), but I incline to hold off for now and revisit it later: I am not sure about the special handling for exclusive access in `shared_library.h`. @deadeyegoodwin I wonder if you have some insight regarding this? I guess there is no way to guarantee it if the backends also interact with shared library.

deadeyegoodwin: Do not make a backend dependent on any internal core library functions, only the TRITONBACKEND and TRITONSERVER functions can be used. As Guan mentioned we may more the utility into common repo at some point, but for now you should just copy what you need into the backend itself, and make sure to change the namespace, etc. to make the copy a full part of the backend.

dyastremsky: Ah okay, so the core library is for the server exclusively, the common one is the one that's shared between repos. That makes sense. Thanks, I'll duplicate the necessary pieces.

src/filesystem.cc
```diff
@@ -0,0 +1,71 @@
...
+    return std::string(".");
+  }
+  if (idx == 0) {
+    return std::string("/");
+  }
+
+  return path.substr(0, idx);
+}
+
+}}}  // namespace triton::backend::tensorrt
```
Tabrizian, line 71: New line

src/shared_library.cc
```diff
@@ -0,0 +1,176 @@
...
+          ("unable to unload shared library: " + std::string(dlerror()))
+              .c_str());
+    }
+#endif
+  }
+
+  return nullptr;  // success
+}
+
+}}}  // namespace triton::backend::tensorrt
```
Tabrizian, line 176: New line

src/shared_library.cc
```diff
@@ -131,7 +131,7 @@ OpenLibraryHandle(const std::string& path, void** handle)
         ("unable to load shared library: " + errstr).c_str());
   }
 #else
-  *handle = dlopen(path.c_str(), RTLD_NOW | RTLD_LOCAL);
+  *handle = dlopen(path.c_str(), RTLD_LAZY);
```
Tabrizian, line 134: Why we need to change the flag here? Can you add a comment explaining the same?

dyastremsky: Thanks, reverted these flags and also addressed your newline comments.

I updated the flags while debugging to see if lazy loading was necessary, but looks not to be, so will use the same flags as we use elsewhere for dlopen again. Rerunning CI to confirm.

tanmayv25 (approved):

PR #26, opened 2022-03-15
Fix to support dynamically batched requests across different optimiza…

tanmayv25:
…tion profiles

Fixes https://github.com/triton-inference-server/server/issues/3852

Tested here: https://github.com/triton-inference-server/server/pull/4056

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -2835,6 +2835,9 @@ ModelInstanceState::EvaluateTensorRTContext(
     for (uint32_t dim_idx = 0; dim_idx < input_dims_count; dim_idx++) {
       input_shape_vec.push_back(*(input_shape + dim_idx));
     }
+    if (support_batching_) {
+      input_shape_vec[0] = total_batch_size;
+    }
 
     int io_index = engine_->getBindingIndex(input_name);
     auto& io_binding_info =


src/tensorrt.cc
```diff
@@ -2835,6 +2835,9 @@ ModelInstanceState::EvaluateTensorRTContext(
     for (uint32_t dim_idx = 0; dim_idx < input_dims_count; dim_idx++) {
       input_shape_vec.push_back(*(input_shape + dim_idx));
     }
+    if (support_batching_) {
+      input_shape_vec[0] = total_batch_size;
```
GuanLuo, line 2839: The vector is the concatenation of all input shape, so if the model has multiple inputs (i.e. input0 [2, 3], input1 [2, 5, 6]), the vector will be [2, 3, 2, 5, 6] and you need to change all 2 to current batch size

tanmayv25: The vector does not appear to be a concatenation of all input shapes. We have a new input_shape_vec for each input.
https://github.com/triton-inference-server/tensorrt_backend/blob/92ae072b4a00d9a08e36d13aee158af65efbb243/src/tensorrt.cc#L2834
Wasn't this expected?

tanmayv25: The error distance is being accumulated from each input separately.

GuanLuo: You are right, I was thinking about how we look up the cuda graphs, your change is correct.

GuanLuo (approved):

src/tensorrt.cc
```diff
@@ -2835,6 +2835,9 @@ ModelInstanceState::EvaluateTensorRTContext(
     for (uint32_t dim_idx = 0; dim_idx < input_dims_count; dim_idx++) {
       input_shape_vec.push_back(*(input_shape + dim_idx));
     }
+    if (support_batching_) {
```
CoderHam, line 2838: Why not handle this inside the above for loop? That way input_shape_vec[0] will be updated only once.

CoderHam (changes requested):

src/tensorrt.cc
tanmayv25: I think it is more clear this way and describe what we are trying to do. Otherwise there will be an additional nesting in for look. A single write will not be that expensive.

CoderHam: Sure

CoderHam (approved):

PR #27, opened 2022-03-17
Fix cuda graph capture mode

tanmayv25:
Follows fixes from the sample file: https://github.com/NVIDIA/TensorRT/blob/main/samples/common/sampleInference.cpp#L666

I have opened [a ticket](https://jirasw.nvidia.com/browse/DLIS-3578) to add extensive testing for cuda graph feature. Tested the outputs from add_sub model locally for now.

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -4945,7 +4945,8 @@ ModelInstanceState::BuildCudaGraph(
     auto buffer_binding_index = num_copy_streams_ == 1 ? 0 : set_idx;
     cudaGraph_t graph;
     // Using cudaStreamCaptureModeThreadLocal mode to confine the graph capture
-    // to local CPU thread.
+    // to this thread and avoid interference from other potentially unsafe cuda
+    // calls.
     auto cuerr =
         cudaStreamBeginCapture(CudaStream(), cudaStreamCaptureModeThreadLocal);
     if (cuerr != cudaSuccess) {
@@ -5088,7 +5089,8 @@ ModelInstanceState::BuildCudaGraphV2(
     cudaGraph_t graph;
     int buffer_bindings_index = num_copy_streams_ == 1 ? 0 : set_idx;
     // Using cudaStreamCaptureModeThreadLocal mode to confine the graph capture
-    // to local CPU thread.
+    // to this thread and avoid interference from other potentially unsafe cuda
+    // calls.
     auto cuerr =
         cudaStreamBeginCapture(CudaStream(), cudaStreamCaptureModeThreadLocal);
     if (cuerr != cudaSuccess) {
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -1097,9 +1097,6 @@ class ModelInstanceState : public TensorRTModelInstance {
       cudaGraphExec_t cuda_graph_exec_;
     };
 
-    // The CUDA graphs captured for the model for different
-    // batch-sizes.
-    std::vector<cudaGraph_t> cuda_graphs_;
     // The key is packed input dimensions prepended by batch size, so
     // that uniqueness is guaranteed and the CUDA graphs are sorted to
     // provide convinence to find the closest CUDA graph in the
@@ -1531,18 +1528,6 @@ ModelInstanceState::~ModelInstanceState()
       }
     }
     trt_context.second.cuda_graph_execs_.clear();
-
-    for (const auto& cuda_graph : trt_context.second.cuda_graphs_) {
-      cudaError_t err = cudaGraphDestroy(cuda_graph);
-      if (err != cudaSuccess) {
-        LOG_MESSAGE(
-            TRITONSERVER_LOG_ERROR,
-            (std::string("Failed to destroy cuda graph: ") +
-             +cudaGetErrorString(err))
-                .c_str());
-      }
-    }
-    trt_context.second.cuda_graphs_.clear();
   }
 
   if (stream_ != nullptr) {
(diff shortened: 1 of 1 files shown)


src/tensorrt.cc
```diff
@@ -5086,6 +5103,24 @@ ModelInstanceState::BuildCudaGraphV2(
...
+        if (cuerr != cudaErrorStreamCaptureInvalidated) {
+          LOG_MESSAGE(
+              TRITONSERVER_LOG_ERROR,
+              (std::string("stream capture is not invalidated for ") + Name() +
+               ": " + cudaGetErrorString(cuerr))
+                  .c_str());
+        }
+        cudaGetLastError();
+      }
+      auto cuerr2 = cudaGetLastError();
```
rmccorm4, line 5116: A comment on the motivation for the extra `cudaGetLastError()` would be useful. Seems like we're stepping back up to find an underlying error, but not totally clear why at a glance.

src/tensorrt.cc
```diff
@@ -5066,7 +5083,7 @@ ModelInstanceState::BuildCudaGraphV2(
     cudaGraph_t graph;
     int buffer_bindings_index = num_copy_streams_ == 1 ? 0 : set_idx;
     auto cuerr =
-        cudaStreamBeginCapture(CudaStream(), cudaStreamCaptureModeGlobal);
+        cudaStreamBeginCapture(CudaStream(), cudaStreamCaptureModeThreadLocal);
```
rmccorm4, line 5086: A comment on the motivation for ThreadLocal mode vs Global mode in the few spots changed would also be useful as a reader

src/tensorrt.cc
```diff
@@ -4194,6 +4175,14 @@ ModelInstanceState::InitializeExecuteInputBinding(
          "' for " + Name() + ": " + cudaGetErrorString(err))
             .c_str());
   }
+  err = cudaMemset((uint8_t*)buffer, 0, max_byte_size);
```
rmccorm4, line 4178: Was this a bug that we weren't previously clearing the buffer? Or leftover artifact from debugging?

src/tensorrt.cc
```diff
@@ -4975,6 +4965,24 @@ ModelInstanceState::BuildCudaGraph(
...
       cuerr = cudaStreamEndCapture(CudaStream(), &graph);
+      if (captured == false) {
+        if (cuerr != cudaErrorStreamCaptureInvalidated) {
+          LOG_MESSAGE(
+              TRITONSERVER_LOG_ERROR,
+              (std::string("stream capture is not invalidated for ") + Name() +
+               ": " + cudaGetErrorString(cuerr))
+                  .c_str());
+        }
+        cudaGetLastError();
```
dyastremsky, line 4976: Is the behavior here correct? It looks like you're getting the last error but not saving it. Then you're grabbing the last error again in a few lines. A little bit hard to understand what's going on.

tanmayv25: I will add the comment. It is just clearing off the error from cuda runtime.

src/tensorrt.cc
tanmayv25: Returns the last error that has been produced by any of the runtime calls in the same host thread and resets it to [cudaSuccess](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038e355f04607d824883b4a50662830d591).
We are just verifying that the state was properly set to cudaSuccess.

src/tensorrt.cc
dyastremsky: Is there a situation when it would not?

src/tensorrt.cc
tanmayv25: This was not the bug in this case but it is a good to have because some models can be data sensitive.

src/tensorrt.cc
tanmayv25: The cudaGetLastError() documentation https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__ERROR.html#group__CUDART__ERROR_1g3529f94cb530a83a76613616782bd233
says that it resets the state to cudaSuccess, but still I see the similar construct in most of TRT cuda graph related code. Hence, keeping it here as well just to add visibility. The server will just print an additional log saying the error was not cleaned.

dyastremsky: Thanks for clarifying.

dyastremsky (approved):

src/tensorrt.cc
rmccorm4: @tanmayv25 to clarify, I meant to ask, why is it important to keep the cuda graph capture local to the thread? Is there some kind of race with multiple model instances of the same TRT cuda graph model? Or even a race across separate loaded TRT cuda graph models?

tanmayv25: The scenario this change fixes is when you already have a model(say for example Model A) hosted on triton at the time this graph capture is initiated(say for Model B). The global setting would fail the graph capture of Model B if there is an uncapturable call anywhere in the system. Some of these APIs might not be tied to a specific stream. Additionally, It is not necessary for Model A to be a cuda graph model. 
That being said we have dedicated threads for all our models, capturing graphs on their unique streams. Hence, making thread_local relaxes the cuda graph capturability tests, allowing successful graph capture and deployment.

tanmayv25: There are lots of places in TRT repo that explains this functionality and I do not want to add all that special cases in the comments in our code base.

rmccorm4: > The global setting would fail the graph capture of Model B if there is an uncapturable call anywhere in the system.

This was more along the lines of the comment I expected for future readers of the code to understand.

Your explanation makes total sense, thanks!

rmccorm4 (approved):
LGTM, thanks for the comments!

GuanLuo (approved):

PR #29, opened 2022-03-18
Update Documentation for Loading Custom Plugins

dyastremsky:
This ticket will update documentation once https://github.com/triton-inference-server/tensorrt_backend/pull/25 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,12 @@ backend that are then applied to all models that use the backend.
 
 ##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>,plugins=/path/plugin1.so;/path2/plugin2.so
 
-Coalesce-request-input instructs TensorRT to consider the requests' inputs with the same name as
+The coalesce-request-input flag instructs TensorRT to consider the requests' inputs with the same name as
 one contiguous buffer if their memory addresses align with each other.
 This option should only be enabled if all requests' input tensors are allocated
 from the same memory region. Default value is false.
 
-Plugins provides a way to load any custom TensorRT plugins that your models rely on. If you have
+The plugins flag provides a way to load any custom TensorRT plugins that your models rely on. If you have
 multiple plugins to load, use a semicolon as the delimiter.
 
 ## Build the TensorRT Backend
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ main Triton [issues page](https://github.com/triton-inference-server/server/issu
 The command-line options configure properties of the TensorRT
 backend that are then applied to all models that use the backend.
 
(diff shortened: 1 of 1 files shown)


dyastremsky:
@tanmayv25 Would "+" be a good delimiter for users to load multiple custom plugins via the backend config?

CoderHam:
I'd recommend ";"

dyastremsky:
Thanks! Easy enough change. :)

README.md
```diff
@@ -50,13 +50,16 @@ main Triton [issues page](https://github.com/triton-inference-server/server/issu
 The command-line options configure properties of the TensorRT
 backend that are then applied to all models that use the backend.
 
-##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>
+##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>,plugins=/path/plugin1.so;/path2/plugin2.so
 
-Instruct TensorRT to consider the requests' inputs with the same name as
+Coalesce-request-input instructs TensorRT to consider the requests' inputs with the same name as
```
tanmayv25, line 55: Make it "coalesce-request-input" with a lower case c.

tanmayv25:
Looks good to me. ";" is usually used for delimiting the paths.

README.md
dyastremsky: Thanks for the feedback. Done.

dyastremsky:
We can't use a semicolon, because the command line will read it as a separate command. The user would then have to use an escape character beforehand, but it seems like our parser turns that into \\;

I think this is why I ended up using a plus sign. Through testing, it looks like a semicolon will work with single quotes (-backend-config='/path/to/plugin;/path2/to/plugin2'), but that deviates from our typical backend-config command behavior. Would you prefer the single quotes or a different delimiter? @tanmayv25 @CoderHam

README.md
```diff
@@ -50,13 +50,16 @@ main Triton [issues page](https://github.com/triton-inference-server/server/issu
 The command-line options configure properties of the TensorRT
 backend that are then applied to all models that use the backend.
 
-##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>
+##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>,plugins=/path/plugin1.so;/path2/plugin2.so
```
CoderHam, line 53: Use quotes around plugin paths if needed as per your earlier comments

CoderHam (commented):
`;` is the native way to separate paths on windows.
`:` is the native way to separate paths on linux and the other candidate to consider. 
I will let you choose which of these you think is better.

I would rather use one of these and ask users to use quotes if needed than use another path separator. 
PS `;` separated list of paths is same way paths are passed in [cmake](https://cmake.org/cmake/help/latest/command/cmake_path.html).

tanmayv25 (approved):

PR #33, opened 2022-05-24
Handle implicit batch dimension for non batching models

tanmayv25:
According to the docs: 

> For an engine built from an [INetworkDefinition](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_network_definition.html) without an implicit batch dimension, this will always return 1.

Source:
https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_cuda_engine.html#a5c2185618a25cc94b588b3318be3b27d

This fixes `nobatch_*` model tests in likes of `L0_trt_reformat_free`.

@nv-kmcgill53  See how a lack of `config_batch_hint` in presence of non-zero `tensors_with_config_shape_cnt`  is treated as a case for batching being disabled.

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -596,7 +596,7 @@ ModelState::AutoCompleteConfigHelper(const std::string& model_path)
   if (engine->hasImplicitBatchDimension()) {
     // If engine has implicit batch dimension then retrieve the value and exit
     max_batch_size = engine->getMaxBatchSize();
-    has_implicit_batch_dim = true;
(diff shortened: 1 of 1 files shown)


nv-kmcgill53 (approved):
So a user which defines at least one IO shape and the batch hint is false, autocomplete is pessimistic about batching the model?

tanmayv25:
Exactly.

PR #34, opened 2022-05-26
Fix implicit state for models that have more than one state

Tabrizian:
We should be adding to the list of sequence states in the payload instead of overriding it. This fixes implicit state for TRT models that have more than one implicit state.

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -2440,10 +2440,13 @@ ModelInstanceState::Run(
       }
 
       if (io_binding_info.is_state_output_) {
-        payload_->seq_states_ = payload_->responder_->ProcessStateTensor(
+        auto updated_states = payload_->responder_->ProcessStateTensor(
             name, dt, batchn_shape,
             static_cast<const char*>(io_binding_info.buffer_),
             io_binding_info.memory_type_, io_binding_info.memory_type_id_);
+        payload_->seq_states_.insert(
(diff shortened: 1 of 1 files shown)


rmccorm4 (approved):
LGTM, logic seems similar to how [ORT backend handles it](https://github.com/triton-inference-server/onnxruntime_backend/blob/e32a32cc323d44710f47c1b66c4ce528a8bdfdb9/src/onnxruntime.cc#L2211-L2227). Though looks like ORT updates states for each output, while TRT gathers all the states from each output and then updates them all at the end.

nv-kmcgill53 (commented):
```
 payload_->seq_states_.insert(
            payload_->seq_states_.end(), updated_states.begin(),
            updated_states.end());
```
Is this a possible memory leak? It's a little hard for me to see the lifetime of `seq_states_`. I guess another question for understanding is: Is the lifetime of the Payload object the same as the incoming request?

rmccorm4:
@nv-kmcgill53 my understanding is `payload_` would be ultimately cleaned up automatically as a unique_ptr, and reset (deleting underlying payload object and clearing states) for each call to [ModelInstanceState::Run](https://github.com/triton-inference-server/tensorrt_backend/blob/4ae952f11d8d08cf7fa2f6722ee446f9878acc20/src/tensorrt.cc#L1674). 

But I'll let Iman confirm.

nv-kmcgill53 (approved):

PR #36, opened 2022-06-15
Set the provided GPU device before trying to load the model for auto-…

tanmayv25:
…complete

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -483,6 +483,42 @@ ModelState::AutoCompleteConfig()
 TRITONSERVER_Error*
 ModelState::AutoCompleteConfigHelper(const std::string& model_path)
 {
+  // Get one of the specified device id from the config to load the model
+  int current_device;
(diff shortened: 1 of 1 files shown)


nv-kmcgill53 (approved):

Tabrizian (commented):
I was wondering how this works if the user uses the `cc_model_filenames` field and multiple GPUs in the instance group setting? Is it possible that in that case we might need to set the CUDA context to a GPU other than first GPU?

tanmayv25:
> I was wondering how this works if the user uses the `cc_model_filenames` field and multiple GPUs in the instance group setting? Is it possible that in that case we might need to set the CUDA context to a GPU other than first GPU?

Good question. I think currently `cc_model_filenames` support is broken for auto-complete config.
https://github.com/triton-inference-server/tensorrt_backend/blob/main/src/tensorrt.cc#L456-L468

I will fix that support for the auto-complete with other PR. It should not be too complicated.

PR #37, opened 2022-06-15
Using cc_model_filename to select which serialized model to use for a…

tanmayv25:
…uto-completing the config

Follow-up PR from #36.


tanmayv25:
Unfortunately, we don't have machines with different GPUs on dlcluster. I need to think how to add these combinations to testing.

nv-kmcgill53:
Could we request dlcluster to construct a heterogeneous machine? I think it's not too big of an ask.

nv-kmcgill53 (approved):

Tabrizian (approved):

PR #39, opened 2022-06-29
Use the SetModelConfig backend utility and parse the config again

tanmayv25:
Uses the utility added by PR: https://github.com/triton-inference-server/backend/pull/64/files

After the update, we run the parsing on entire config just to be more future-proof in case backends add more auto-complete functionalities.

### DIFF
--- a/src/tensorrt_model.cc
+++ b/src/tensorrt_model.cc
@@ -58,6 +58,15 @@ TensorRTModel::TensorRTModel(TRITONBACKEND_Model* triton_model)
   ParseModelConfig();
 }
 
+TRITONSERVER_Error*
+TensorRTModel::SetTensorRTModelConfig()
+{
+  RETURN_IF_ERROR(SetModelConfig());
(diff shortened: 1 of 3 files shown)


krishung5 (approved):

nv-kmcgill53 (approved):

Tabrizian (approved):

src/tensorrt_model.cc
```diff
@@ -58,6 +58,15 @@ TensorRTModel::TensorRTModel(TRITONBACKEND_Model* triton_model)
   ParseModelConfig();
 }
 
+TRITONSERVER_Error*
+TensorRTModel::SetTensorRTModelConfig()
+{
+  RETURN_IF_ERROR(SetModelConfig());
+  ParseModelConfig();
```
Tabrizian, line 65: It looks like SetModelConfig calls `ParseModelConfig` too so perhaps we can remove the additional `ParseModelConfig` here?

PR #42, opened 2022-07-12
Fix handling of bs 1 with batching enabled in implicit mode

tanmayv25:


src/tensorrt.cc
```diff
@@ -670,7 +670,7 @@ ModelState::AutoCompleteConfigHelper(const std::string& model_path)
   if (engine->hasImplicitBatchDimension()) {
     // If engine has implicit batch dimension then retrieve the value and exit
     max_batch_size = engine->getMaxBatchSize();
-    has_implicit_batch_dim = (max_batch_size != 1);
+    has_implicit_batch_dim = (max_batch_size != 1) || (MaxBatchSize() != 0);
```
rmccorm4, line 673: Is the Triton config specifying implicit batch enough to make it work on the engine in the backend even if `engine->getMaxBatchSize()` disagrees? Is that logic handled?

tanmayv25: I think the thing you are asking is being handled here: https://github.com/triton-inference-server/tensorrt_backend/blob/tanmayv-fix/src/tensorrt.cc#L700-L714

tanmayv25: | MBS within TRT Engine | MBS provided by Triton  | Result |
|---|---|---|---|---|
|   1| 1  |  Success  | 
|   1|  10 |  Fail | 
|   1|  100 | Fail  |

nv-kmcgill53: Is there a case where the `engine->hasImplicitBatchDimension() == true` and `engine->getMaxBatchSize() == 0` ? If so, should L673 become `has_implicit_batch_dim = (max_batch_size > 1) || (MaxBatchSize() != 0);`?

tanmayv25: `engine->getMaxBatchSize() == 0`
This is never true. getMaxBatchSize API  returns value greater than or equal to 1. 
Note this is max_batch_size from the TensorRT engine and is different from the Triton's max_batch_size.

nv-kmcgill53 (approved):

src/tensorrt.cc
tanmayv25: The difference is coming from how Triton and TensorRT defines the batch size differently. The max_batch_size of 0 in triton means that batching is not available but Triton can still forward complete tensors to the TensorRT engine. The TensorRT engine on the other hand does not have any concept of max_batch_size=0.

The presence of implicit batch size is used in auto-complete logic to detect whether or not to set batching in the model. From Triton's  perspective, TensorRT engine supports implicit batching iff the max_batch_size within engine is greater than or equal to 1 or user has explicitly provided some non-zero value within config.pbtxt.

We will still use enqueue_v1. May be the variable name can be modified to avoid confusion as it is Triton's interpretation of whether engine supports batching or not.

PR #44, opened 2022-07-25
Add RETURN_IF_ERROR for allow_ragged_batch variable

krishung5:
We need to handle the returned `TRITONSERVER_ErrorNew` object properly to prevent memory leak from unhandled `TRITONSERVER_ErrorNew`.
I noticed in our repos, there are some APIs that are not being handled for this properly, i.e. missing `RETURN_IF_ERROR` for API calls. Created a follow-up ticket DLIS-4007 for checking if we handle error return for all the TritonJson API calls.
Related PRs:
server: https://github.com/triton-inference-server/server/pull/4693
common: https://github.com/triton-inference-server/common/pull/67
backend: https://github.com/triton-inference-server/backend/pull/67

### DIFF
--- a/src/tensorrt.cc
+++ b/src/tensorrt.cc
@@ -631,7 +631,10 @@ ModelState::AutoCompleteConfigHelper(const std::string& model_path)
            ((i < config_io.ArraySize()) && (!io_allow_ragged_batch)); i++) {
         triton::common::TritonJson::Value io;
         RETURN_IF_ERROR(config_io.IndexAsObject(i, &io));
-        io.MemberAsBool("allow_ragged_batch", &io_allow_ragged_batch);
+        triton::common::TritonJson::Value allow_ragged_batch;
+        if (io.Find("allow_ragged_batch", &allow_ragged_batch)) {
+          RETURN_IF_ERROR(allow_ragged_batch.AsBool(&io_allow_ragged_batch));
+        }
         if (io_allow_ragged_batch) {
           // Treat the presence of tensor allowing ragged batch as
           // a hint for batching.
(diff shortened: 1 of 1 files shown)


tanmayv25 (approved):

tanmayv25:
You are right. We keep digressing and introducing the memory leaks because of unhandled objects. Thanks for creating the ticket.

src/tensorrt.cc
```diff
@@ -631,7 +631,10 @@ ModelState::AutoCompleteConfigHelper(const std::string& model_path)
            ((i < config_io.ArraySize()) && (!io_allow_ragged_batch)); i++) {
         triton::common::TritonJson::Value io;
         RETURN_IF_ERROR(config_io.IndexAsObject(i, &io));
-        io.MemberAsBool("allow_ragged_batch", &io_allow_ragged_batch);
+        triton::common::TritonJson::Value allow_ragged_batch;
+        if (io.Find("allow_ragged_batch", &allow_ragged_batch)) {
+          RETURN_IF_ERROR(allow_ragged_batch.AsBool(&io_allow_ragged_batch));
```
GuanLuo, line 636: @tanmayv25 Do we want to abort the auto complete if "allow_ragged_batch" is not found?

GuanLuo: We can probably change the original line to below to delete error message if returned.
```
auto err = io.MemberAsBool("allow_ragged_batch", &io_allow_ragged_batch);
if (err) { ErrorDelete(err); }
```

krishung5: I was wondering why we can abort the auto-complete if "allow_ragged_batch" is not found? I thought "allow_ragged_batch" is not a must-have field for model configuration. We might still need to run the auto-complete logic for max_batch_size and inputs/outputs?

GuanLuo: Sorry I read it wrong, I was thinking it will return error and thus return early if "allow_ragged_batch" is not specified.

tanmayv25: @GuanLuo  We can do this to save multiple lines or clearer logic, but do note that current implementation is not aborting the auto-complete if  "allow_ragged_batch" is not found. It is aborting if it is unable to read "allow_ragged_batch", if present, as boolean value.

PR #46, opened 2022-09-09
ifdef out kUINT8 based on TRT version

krishung5:


nv-kmcgill53 (approved):
Generated ticket specifying criteria for removal of the ifdefs: DLIS-4144

src/tensorrt_utils.cc
```diff
@@ -25,11 +25,16 @@
...
-
+#include <NvInferPlugin.h>
 #include "triton/backend/backend_common.h"
 
 namespace triton { namespace backend { namespace tensorrt {
 
+// Cannot directly use 'NV_TENSORRT_VERSION' in preprocessor because of the
+// syntax.
+int32_t version_ = NV_TENSORRT_VERSION;
+#define TENSORRT_VERSION version
```
GuanLuo, line 36: Have you verified your change? "_" is missing here

krishung5: Sorry for missing that. Added `_` last minute and somehow it's still compiled. It's fixed now.

src/tensorrt_utils.cc
```diff
@@ -25,11 +25,16 @@
...
 
 #include "tensorrt_utils.h"
-
+#include <NvInferPlugin.h>
 #include "triton/backend/backend_common.h"
 
 namespace triton { namespace backend { namespace tensorrt {
 
+// Cannot directly use 'NV_TENSORRT_VERSION' in preprocessor because of the
+// syntax.
```
GuanLuo, line 34: Why can't be used? Because `NV_TENSORRT_VERSION` is defined as `constexpr`? Can you use https://github.com/NVIDIA/TensorRT/blob/main/include/NvInferVersion.h#L22-L23 ?

src/tensorrt_utils.cc
rmccorm4: > Added _ last minute and somehow it's still compiled.

FYI, the definition won't do any checking. It will only have a compiler error if you tried to substitute the variable somewhere outside of an ifdef / directive.

```
#include <iostream>
#define MY_VAR non_existent_var

int main() {
    #if MY_VAR > 123
      std::cout << "Set correctly as some number";
    #else
      std::cout << "Set incorrectly";  // OK
      std::cout << "Set incorrectly as: " << MY_VAR;  // COMPILER ERROR UNDEFINED VARIABLE
    #endif

    return 0;
}
```

src/tensorrt_utils.cc
krishung5: It's because the `::` in this line: https://github.com/NVIDIA/TensorRT/blob/b55c4710ce01f076c26710a48879fcb2661be4a9/include/NvInferRuntimeCommon.h#L73

Used `NV_TENSORRT_MAJOR` and `NV_TENSORRT_MINOR `.

src/tensorrt_utils.cc
```diff
@@ -45,10 +40,10 @@ ConvertTrtTypeToDataType(nvinfer1::DataType trt_type)
       return TRITONSERVER_TYPE_FP16;
     case nvinfer1::DataType::kINT8:
       return TRITONSERVER_TYPE_INT8;
-#if (TENSORRT_VERSION >= 8500)
+#if ((NV_TENSORRT_MAJOR >= 8) && (NV_TENSORRT_MINOR >= 5))
```
rmccorm4, line 43: this will break on TRT 9.0 if we don't remove it before then

krishung5: Ah right.. Let me change the logic here.

krishung5: Would simply use `[NV_TENSORRT_VERSION](https://github.com/NVIDIA/TensorRT/blob/b55c4710ce01f076c26710a48879fcb2661be4a9/include/NvInferRuntimeCommon.h#L82-L83)` be a better solution? It calculates the version and we can just compare it to the number "8500"?

nv-kmcgill53: I would be in favor of `NV_TENSORRT_VERSION`.

GuanLuo: `NV_TENSORRT_VERSION` is `constexpr`, I am not experience with it so my question is can it replace preprocessor directive to control which part of the code to be hidden for compilation? The issue is that `kUINT8` is not available for older version so we don't want the compiler to do anything with that

krishung5: Sorry I don't follow what you mean.. We can use `NV_TENSORRT_VERSION` to determine if `kUINT8` should be hidden for compilation or not. The issue I'm facing is just that we cannot directly use `NV_TENSORRT_VERSION` like this: `#if (NV_TENSORRT_VERSION  >= 8500)` because colons are not acceptable in `#if` directive. We would need to have some workaround as the previous commit. I don't think whether `NV_TENSORRT_VERSION` is `constexpr` or not would affect anything here.

GuanLuo: Okay, just make sure it can build with older TRT version (22.08 container)

krishung5:
Reverted back to the previous commit which uses `NV_TENSORRT_VERSION`. I build TRT backend with 22.08 container locally and it's working. Waiting on the CI.

nv-kmcgill53 (approved):

src/tensorrt_utils.cc
rmccorm4: @krishung5 I think what Guan was asking is if something like this would compile with TRT < 8.5 using constexpr instead of ifdef, since constexpr is a compile-time concept.

(Could use the TRT constexpr variable instead of mine in this example)

```cpp
constexpr bool has_uint8 = false;

if (has_uint8) {
  dtype = nvinfer::dtype::kuint8
}
```

However, I think the answer is no, I believe the compiler would still complain the type doesn't exist for TRT < 8.5 even if using the constexpr.

krishung5: I see. Thanks for the clarification!

nv-kmcgill53: You are right that using a constexpr doesn't solve the issue. [Here is an example where the preprocessor allows the code to compile but the `constexpr` doesn't](https://godbolt.org/z/MbqTEPWzE). It has to do with when the compiler does the type checking, in the case of the C++ compiler it's done after the preprocessor. So any macros which are defined get substituted in as if it was explicitly written before compilation happens.

nv-kmcgill53 (changes requested):
This change fails when using [TRT_VERSION 8.5.0.12](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/44235796) but succeeds with [TRT_VERSION 8.4.2.4](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/44232642).

nv-kmcgill53:
Use this as the #define

`#define TENSORRT_VERSION ((NV_TENSORRT_MAJOR * 1000) + (NV_TENSORRT_MINOR * 100) + NV_TENSORRT_PATCH)`

nv-kmcgill53 (approved):

PR #54, opened 2022-11-03
Support data dependent shape output. Use TRT v3 APIs

GuanLuo:


src/instance_state.cc
```diff
@@ -3913,24 +3898,42 @@ TRTv1Interface::SetBindingDimensions(
...
 bool
-TRTv2Interface::Enqueue(nvinfer1::IExecutionContext* context)
+TRTv3Interface::Enqueue(nvinfer1::IExecutionContext* context)
 {
-  return context->enqueueV2(
-              instance_->buffer_bindings_[instance_->next_buffer_binding_set_].data(), instance_->stream_,
-              &instance_->events_[instance_->next_set_].ready_for_input_);
+  if (SetTensorAddress(context)) {
+    if (context->setInputConsumedEvent(instance_->events_[instance_->next_set_].ready_for_input_)) {
+      return context->enqueueV3(instance_->stream_);
```
tanmayv25, line 3905: This looks much cleaner!

tanmayv25 (approved):
I believe we haven't tested these changes on jetson because trt 8.5 is not available on jetpack.
We might have to look into the zero copy support on jetson properly.

Tabrizian (approved):

GuanLuo:
I don't think there will be change in terms of Jetson's "zero copy" support, but there is opportunity for simplifying the handling as TRT 8.5. simplifies I/O tensor placement (memory location) and provides API to query the information which we can take advantage to unify binding initialization between shape tensor and execution tensor.

PR #59, opened 2022-12-14
Add virtual destructor to the base class

krishung5:
The memory leak happened because the object in the derived class was not cleaned up. Added virtual destructor to the base class so that the destructor of derived class would be called.


tanmayv25 (approved):

rmccorm4 (approved):

nv-kmcgill53 (approved):
Tests look good.

PR #67, opened 2023-04-17
Add more context to TensorRT error message

kthui:
Currently, the TensorRT error message are only printed to Triton error log, but not cannot be propagated back to the Triton client making the request. The PR adds the capability to store the last TensorRT error message from the model that generates the error, and propagates the error back to the Triton client, as a supplement for existing error messages that lack some context (i.e. "unable to create TensorRT runtime" and "unable to create TensorRT context")

Related PR: https://github.com/triton-inference-server/server/pull/5651
Closes triton-inference-server/server/issues/5304

### DIFF
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -35,6 +35,7 @@ TensorRTLogger::log(Severity severity, const char* msg) noexcept
 {
   switch (severity) {
     case Severity::kINTERNAL_ERROR:
+      RecordErrorMsg(msg);
       LOG_MESSAGE(TRITONSERVER_LOG_ERROR, msg);
       break;
     case Severity::kERROR:
--- a/src/logging.cc
+++ b/src/logging.cc
@@ -1,4 +1,4 @@
-// Copyright 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -30,8 +30,6 @@
 
 namespace triton { namespace backend { namespace tensorrt {
 
-TensorRTLogger tensorrt_logger;
-
 void
 TensorRTLogger::log(Severity severity, const char* msg) noexcept
 {
@@ -40,6 +38,7 @@ TensorRTLogger::log(Severity severity, const char* msg) noexcept
       LOG_MESSAGE(TRITONSERVER_LOG_ERROR, msg);
       break;
     case Severity::kERROR:
+      StoreMsg(severity, msg);
       LOG_MESSAGE(TRITONSERVER_LOG_ERROR, msg);
       break;
     case Severity::kWARNING:
@@ -52,6 +51,25 @@ TensorRTLogger::log(Severity severity, const char* msg) noexcept
       LOG_MESSAGE(TRITONSERVER_LOG_VERBOSE, msg);
       break;
   }
+}
+
+void
+TensorRTLogger::StoreMsg(Severity severity, const char* msg) noexcept
+{
+  // Currently, only the 'Severity::kERROR' message is interested and this
+  // function will only be called when a 'Severity::kERROR' message arrives, but
+  // the interface is designed to be called by 'TensorRTLogger::log()' for any
+  // messages under any severity level, if the other messages become interested
+  // in the future.
+  std::lock_guard<std::mutex> lock(mu_);
+  last_error_msg_ = std::string(msg);
+}
+
+std::string
+TensorRTLogger::LastErrorMsg()
+{
+  std::lock_guard<std::mutex> lock(mu_);
+  return last_error_msg_;
 }
 
 }}}  // namespace triton::backend::tensorrt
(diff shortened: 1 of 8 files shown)


src/logging.h
```diff
@@ -27,13 +27,26 @@
...
-};
 
-extern TensorRTLogger tensorrt_logger;
+ public:
+  // Return the last 'Severity::kERROR' message logged by TensorRT
+  std::string LastErrorMsg();
+
+ private:
+  // Store messages logged by TensorRT
+  void StoreMsg(Severity severity, const char* msg) noexcept;
```
tanmayv25, line 46: *RecordErrorMsg

src/logging.h
```diff
@@ -27,13 +27,26 @@
...
+
+  std::mutex mu_;  // serialize all state changing operations
```
tanmayv25, line 48: *last_error_msg_mu_

tanmayv25: update the comment above to // Protect the last error message string.

src/logging.cc
```diff
@@ -54,4 +53,23 @@ TensorRTLogger::log(Severity severity, const char* msg) noexcept
   }
 }
 
+void
+TensorRTLogger::StoreMsg(Severity severity, const char* msg) noexcept
+{
+  // Currently, only the 'Severity::kERROR' message is interested and this
```
tanmayv25, line 59: What about kINTERNAL_ERROR severity? Shouldn't we record and report these details as well. Can you read about it when and where these might happen?

src/logging.h
kthui: Update commit: [Refactor for more specificity](https://github.com/triton-inference-server/tensorrt_backend/pull/67/commits/2fc9058874e7efb2ca7e297ac4d676dcb0ccbff2)

src/logging.h
kthui: Update commit: [Refactor for more specificity](https://github.com/triton-inference-server/tensorrt_backend/pull/67/commits/2fc9058874e7efb2ca7e297ac4d676dcb0ccbff2)

src/logging.cc
kthui: I do not think `Severity::kINTERNAL_ERROR` severity is happening for the case we are addressing. For example, from the error log provided by the user:
```
I0125 20:05:23.645607 87 [logging.cc:49](http://logging.cc:49/)] [MemUsageChange] Init cuDNN: CPU +0, GPU +8, now: CPU 1459, GPU 22254 (MiB)
E0125 20:05:23.696285 87 [logging.cc:43](http://logging.cc:43/)] 1: [defaultAllocator.cpp::allocate::20] Error Code 1: Cuda Runtime (out of memory)
W0125 20:05:23.696333 87 [logging.cc:46](http://logging.cc:46/)] Requested amount of GPU memory (513802240 bytes) could not be allocated. There may not be enough free memory for allocation to succeed.
E0125 20:05:23.720471 87 [logging.cc:43](http://logging.cc:43/)] 2: [executionContext.cpp::ExecutionContext::409] Error Code 2: OutOfMemory (no further information)
I0125 20:05:23.720523 87 [tensorrt.cc:298](http://tensorrt.cc:298/)] TRITONBACKEND_ModelInstanceFinalize: delete instance state
I0125 20:05:23.720547 87 [tensorrt.cc:237](http://tensorrt.cc:237/)] TRITONBACKEND_ModelFinalize: delete model state
E0125 20:05:23.720584 87 model_lifecycle.cc:597] failed to load '495cff3d03904752de92203d58f75f33' version 1: Internal: unable to create TensorRT context
I0125 20:05:23.720695 87 model_repository_manager.cc:1409] Model repository unregistered: /opt/ml/models/495cff3d03904752de92203d58f75f33/model
```
All the error loggings (lines start with an `E`) are at line `logging.cc:43` which is `Severity::kERROR` and non of them are from `Severity::kINTERNAL_ERROR` (which would be at line `logging.cc:40`).

I think we can add it to align with the existing mapping from TensorRT error to Triton error.

Update commit: [Add error msg logging for internal error](https://github.com/triton-inference-server/tensorrt_backend/pull/67/commits/f77a1c9af5920acda919b5620130d00892026348)

tanmayv25: > I do not think Severity::kINTERNAL_ERROR severity is happening for the case we are addressing

I know, but let's try to make it future-proof.

tanmayv25 (dismissed):

src/logging.cc
```diff
@@ -30,16 +30,16 @@
...
 namespace triton { namespace backend { namespace tensorrt {
 
-TensorRTLogger tensorrt_logger;
-
 void
 TensorRTLogger::log(Severity severity, const char* msg) noexcept
 {
   switch (severity) {
     case Severity::kINTERNAL_ERROR:
+      RecordErrorMsg(msg);
```
tanmayv25, line 38: Why not use use switch-case fall-through here? 
The case  implementation is completely identical. We can just remove everything from `Severity::kINTERNAL_ERROR` block and let it pass to the `Severity::kERROR` block below.

kthui: Good idea!
Updated: [Simplify by using switch fall through](https://github.com/triton-inference-server/tensorrt_backend/pull/67/commits/c7cdf3c6fdfc27e5c03dacc2e8e6cd0e95b2737a)

tanmayv25 (approved):

PR #72, opened 2023-07-05
Support for data-dependent output shapes

dyastremsky:
This pull request uses OutputAllocators to handle data-dependent output shapes.

Related server PR: https://github.com/triton-inference-server/server/pull/6057


src/instance_state.h
```diff
@@ -551,6 +554,10 @@ class ModelInstanceState : public TensorRTModelInstance {
 
     // Indicates whether the output is a output tensor.
     bool is_requested_output_tensor_;
+
+    // Whether this is an output using OutputAllocator for
+    // dynamic resizing.
+    bool is_dynamic_{false};
```
tanmayv25, line 560: is_dynamic_ -> is_dynamic_shape_output_ ?

src/output_allocator.cc
```diff
@@ -0,0 +1,75 @@
...
+    output_size_ = 0;
+    if (zero_copy_support_) {
+      cudaHostAlloc(&output_ptr_, size, cudaHostAllocMapped);
+    } else {
+      cudaMalloc(&output_ptr_, size);
+    }
+    // If zero copy support is enabled, need to set the buffer to the device
+    // pointer.
+    if (zero_copy_support_) {
+      void* device_buffer;
```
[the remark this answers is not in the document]
dyastremsky: Good find. Done!

src/output_allocator.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "output_allocator.h"
+
+namespace triton { namespace backend { namespace tensorrt {
+
+void*
+OutputAllocator::reallocateOutput(
+    char const* tensor_name, void* current_memory, uint64_t size,
+    uint64_t alignment) noexcept
+{
+  if (size > output_size_) {
```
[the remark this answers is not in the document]
dyastremsky: Sure, added.

src/instance_state.h
dyastremsky: Updated!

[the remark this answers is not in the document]
dyastremsky: Sure. I've updated the member variable initialization across the file. Guan linked [this](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-in-class-initializer) and wanted to start migrating us toward in-class initializers, though there were concerns of consistency as we started. So at least this file will use in-class initializers, and we can start updating this as we go.

src/instance_state.cc
```diff
@@ -1114,20 +1122,25 @@ ModelInstanceState::Run(
             "failed to run TRT response");
       }
 
+      void* buffer =
```
tanmayv25, line 1125: Can you modify the io_binding_info struct into a class with some of the this logic exposed as function? The abstraction will also reduce code duplicity.

dyastremsky: Good thinking. Done.

src/instance_state.h
```diff
@@ -570,32 +520,35 @@ class ModelInstanceState : public TensorRTModelInstance {
...
+  std::unique_ptr<TRTInterface> interface_{nullptr};
 
   // TRT model instance performs execution asynchorously and thus may go
   // ahead to prepare further executions. Use semaphore to prevent going too
   // far ahead and overwriting resources that are still in use.
-  std::unique_ptr<Semaphore> semaphore_;
+  std::unique_ptr<Semaphore> semaphore_{nullptr};
+
+  // Vector of OutputAllocators to hold pointers to prevent deallocation.
+  std::vector<std::unique_ptr<OutputAllocator>> allocators_{};
```
tanmayv25, line 551: Shouldn't we move the allocators within IOBindingInfo? Each IOBindingInfo will map to a unique OutputAllocator if having dynamic output.

src/io_binding_info.h
```diff
@@ -0,0 +1,148 @@
...
+ public:
+  // Setters and Getters
+  void SetName(const std::string& name) { name_ = name; }
+  const std::string& GetName() const { return name_; }
+
+  void SetByteSize(uint64_t byte_size) { byte_size_ = byte_size; }
+  uint64_t GetByteSize() const { return byte_size_; }
+
+  void SetBuffer(void* buffer) { buffer_ = buffer; }
+  void* GetBuffer() const { return buffer_; }
```
tanmayv25, line 77: With OutputAllocator inside this class, the GetBuffer function can optionally use OutputAllocator or the raw buffer pointer. 
This will help us abstract away OutputAllocator logic from the instance_state class. I believe the only exposure would be SetTensorAddress or SetOutputAllocator calls.

src/instance_state.h
dyastremsky: Good idea. Done.

src/io_binding_info.h
dyastremsky: Great idea. Done.

dyastremsky:
@tanmayv25 This ticket is now ready for another review. I have incorporated your changes.

I think we should review and merge this ticket without the related testing ticket. I tested this locally and the fix works. I am also running a CI pipeline to ensure everything passes. The reason we should do the testing ticket separately is because the new shape tensor tests I added cannot be generated for testing until the next TensorRT container is released.

While we could split that testing PR into two PRs and merge the non-shape testing pieces, the next version of TensorRT should include the fix for the unexpected behavior of getting 0 for max byte size. As a result, we can't replicate that unexpected behavior for the next TRT version and therefore there is no point to switching to using two nonzero op layers in the test anyway. In short, once the new TRT container can be used as an upstream, we can then update the tests to work with the new TRT version.

tanmayv25:
I understand. Let's merge the PR and we can add testing once we move to TRT 9.0

src/output_allocator.cc
```diff
@@ -0,0 +1,77 @@
...
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "output_allocator.h"
+
+namespace triton { namespace backend { namespace tensorrt {
+
+void*
+OutputAllocator::reallocateOutput(
```
Tabrizian, line 32:
```suggestion
OutputAllocator::ReallocateOutput(
```

src/output_allocator.h
```diff
@@ -0,0 +1,68 @@
...
+  // Allocates output dimensions
+  void* reallocateOutput(
+      char const* tensor_name, void* current_memory, uint64_t size,
+      uint64_t alignment) noexcept override;
+
+  // Updates output dimensions
+  void notifyShape(
+      char const* tensor_name, nvinfer1::Dims const& dims) noexcept override;
+
+  void* getBuffer() { return output_ptr_; };
```
Tabrizian, line 49:
```suggestion
  void* GetBuffer() { return output_ptr_; };
```

src/output_allocator.h
```diff
@@ -0,0 +1,68 @@
...
+  void** getBufferAddr() { return &output_ptr_; };
```
Tabrizian, line 50:
```suggestion
  void** GetBufferAddr() { return &output_ptr_; };
```

src/output_allocator.cc
dyastremsky: We can't make this change, as this is implementing the TensorRT IOutputAllocator interface (see [here](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#dynamic-shaped-output)). I think the only function we can make match out code style is GetBuffer() and GetBufferAddr(), but then the function names within the class would be inconsistent.

Tabrizian: got it, thanks for the clarification.

dyastremsky:
Added a few quality-of-life updates. Fixed the test failure (the output allocator was getting created multiple times, so all but the last context would have had their allocator go out of scope): https://github.com/triton-inference-server/tensorrt_backend/pull/72/commits/b3705deeccc4bf2a589beb9bbbc080f4e0264cb0

L0_trt_data_dependent_shape and L0_trt_dynamic_shape passed locally. Re-running CI now.

src/output_allocator.h
```diff
@@ -0,0 +1,68 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <NvInfer.h>
+#include <malloc.h>
+
+namespace triton { namespace backend { namespace tensorrt {
+
+class OutputAllocator : public nvinfer1::IOutputAllocator {
```
tanmayv25, line 34: nit: Can you add a comment here about the purpose of this class and how it extends nvinfer1::IOutputAllocator?
It would avoid confusion like @Tabrizian ran into below.

tanmayv25 (dismissed):

tanmayv25 (approved):

PR #73, opened 2023-07-20
Fix printing instance name when model name is expected

kthui:
The utility function expects a model name, but caller supplied model instance name instead. This PR fixes this problem. See the first parameter of `CompareDimsSupported()`.

The ultimate goal of this PR is to fix L0_model_config, which the model instance name no longer matches the model name when `count = 1` after merging https://github.com/triton-inference-server/core/pull/231.


kthui:
The ultimate goal of this PR is to fix L0_model_config, which the model instance name no longer matches the model name when `count = 1` after merging https://github.com/triton-inference-server/core/pull/231.

rmccorm4 (approved):

PR #74, opened 2023-08-25
Support version compatibility for TensorRT models

dyastremsky:
Adds boolean backend parameter "version-compatible" to allow users to specify whether [version-compatible TensorRT models](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#version-compat) will be supported by Triton.

This pull request had local testing, but I am not adding formalized testing. This is because using version-compatible models in this way is memory-inefficient. Users will likely want to be able to [manually load runtimes](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#manual-load-runtime) to avoid needing to duplicate the lean runtime included in version-compatible models. In the meantime, this is still valuable for users who want to use a small number of version-compatible TensorRT models with Triton.

I am going to work on the runtime reuse design and do the work as a separate ticket, since it modifies the way this backend tracks/uses runtimes. After completing that work, it'll be clearer what testing should look like. Originally, it seemed convenient for CI/CD purposes to have this integrated for generating and loading models used in functional testing. However, given the runtime duplication issue, I won't be able to know whether that approach or adding new models/tests specific to this feature makes more sense until I design the feature to reuse runtimes.

**Testing (by adding `config.set_flag(trt.BuilderFlag.VERSION_COMPATIBLE)` to the model generation, then loading a model and running inference).**
Current main branch:
image

After changes:
image
image

Related tests added to server: https://github.com/triton-inference-server/server/pull/6255

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,9 +72,12 @@ for detail). Currently the following values are accepted:
 * The `plugins` flag provides a way to load any custom TensorRT plugins that your models rely on. If you have
 multiple plugins to load, use a semicolon as the delimiter.
 
-* The `version-compatible` flag enables the loading of version-compatible TensorRT models. You must trust the
-models loaded in this mode, as version-compatible models include a lean runtime which gets deserialized and executed
-by Triton. The default value is false.
+* The `version-compatible` flag enables the loading of version-compatible TensorRT models where the version of
+TensorRT used for building does not matching the engine version used by Triton. You must trust the models
+loaded in this mode, as version-compatible models include a lean runtime which gets deserialized and executed
+by Triton. You can find more information in the TensorRT documentation
+[here](https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#version-compat).
+The default value is false.
 
 
 
--- a/README.md
+++ b/README.md
@@ -72,9 +72,9 @@ for detail). Currently the following values are accepted:
 * The `plugins` flag provides a way to load any custom TensorRT plugins that your models rely on. If you have
 multiple plugins to load, use a semicolon as the delimiter.
 
-* The `version-compatible` flag enables the loading of version-compatible TensorRT models. This is optional,
-as you must trust the models loaded in this mode. Version-compatible models include a lean runtime which gets
-deserialized and executed by Triton. The default value is false.
+* The `version-compatible` flag enables the loading of version-compatible TensorRT models. You must trust the
+models loaded in this mode, as version-compatible models include a lean runtime which gets deserialized and executed
+by Triton. The default value is false.
 
 
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ backend that are then applied to all models that use the backend.
 Below is an example of how to specify the backend config and the full list of
 options.
 
-##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>,plugins="/path/plugin1.so;/path2/plugin2.so"
+##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>,plugins="/path/plugin1.so;/path2/plugin2.so,version-compatible=true"
 
 * The `coalesce-request-input` flag instructs TensorRT to consider the requests' inputs with the same name as
 one contiguous buffer if their memory addresses align with each other.
--- a/README.md
+++ b/README.md
@@ -55,15 +55,12 @@ options.
 
 ##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>,plugins="/path/plugin1.so;/path2/plugin2.so"
 
-* `coalesce-request-input` flag instructs TensorRT to consider the requests' inputs with the same name as
+* The `coalesce-request-input` flag instructs TensorRT to consider the requests' inputs with the same name as
 one contiguous buffer if their memory addresses align with each other.
 This option should only be enabled if all requests' input tensors are allocated
-from the same memory region. Default value is false.
+from the same memory region. The default value is false.
 
-* `plugins` flag provides a way to load any custom TensorRT plugins that your models rely on. If you have
-multiple plugins to load, use a semicolon as the delimiter.
-
-* `execution-policy` flag instructs TensorRT backend to execute the model with
+* The `execution-policy` flag instructs TensorRT backend to execute the model with
 different Triton execution policies (see `TRITONBACKEND_ExecutionPolicy`
 for detail). Currently the following values are accepted:
   * `DEVICE_BLOCKING`: corresponds to `TRITONBACKEND_EXECUTION_DEVICE_BLOCKING`,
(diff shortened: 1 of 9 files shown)


src/loader.cc
```diff
@@ -52,6 +52,14 @@ LoadPlan(
            tensorrt_logger->LastErrorMsg())
               .c_str());
     }
+
+    if (ModelState::isVersionCompatible() &&
+        !runtime->get()->getEngineHostCodeAllowed()) {
```
tanmayv25, line 57: Why do we need `getEngineHostCodeAllowed()` check?
Can't we directly call `setEngineCodeAllowed(ModelState::isVersionCompatible())`?

src/model_state.h
```diff
@@ -134,6 +137,9 @@ class ModelState : public TensorRTModel {
   bool engine_sharing_;
 
   std::unique_ptr<ExecutionArbitrator> execution_arbitrator_;
+
+  // Whether the backend should support version-compatible TensorRT models.
+  static inline bool is_version_compatible_{false};
```
tanmayv25, line 142: Why do we need this variable to be `static`? I wonder if we can have a use-case where we would want to load some models in version compatibility mode enabled and others with version compatibility disabled.

tanmayv25 (commented):
We would need to include some tests for these changes.

src/loader.cc
dyastremsky: I thought it was important that we logged that version compatibility was enabled (especially since invalid backend parameter keys don't return an error), but wanted to make sure it only happened exactly once per-runtime to avoid user confusion or redundant logging. I was trying to think ahead to the next part of this feature where there would be multiple runtimes supported.

One alternate approach could be logging that the version compatibility backend parameter was received. That would also mean that the version compatibility logging statement would happen only once, which I think is better. I can switch to that approach, if that works for you.

src/model_state.h
dyastremsky: I was doing this because the ModelState::CreateEngine call is static, but I realized that the logic made a lot more sense to put inside the logger's LoadPlan function. I think it would still be good to have this be static if we are doing this on a per-server basis.

As far as that use case, maybe. The main reason to disable version compatibility is for security. Doing it on a per-server basis makes sense to me. It's possible a user would have a use case though where they'd want us to allow the setting to be overridden from the model config. It seems like that's a security paradox though. You're essentially saying "I trust the user(s) loading models enough to know which models are safe, but I don't trust the user's model files to all be safe." That may be a long-term ask and doing it now could help with not worrying about backward compatibility later, though it might be premature as users might not want that or have that edge case. If you would rather me add that functionality before this PR gets merged, I'm happy to start a design. The reason for the small design is that I want to think through the best way to do this (e.g. will the backend need to have compatibility enabled then each model has to specifically enable it? what are the defaults?) and keep everything secure.

dyastremsky:
> Can you elaborate on this duplication issue?

By runtime duplication issue above, I mean that if I enabled version compatibility for our QA models, each model would have its own runtime and would thereby bloat the test size. After runtime reusability is added, it seems like more trouble than its worth to generate/save a runtime and adds its path to each of the non-performance test models' configurations. The benefit is that we could not worry about compatibility during the short period between releases when testing. I think it'd make more sense to generate new models and test version compatibility as a unit test. But perhaps I'm missing something that will change that tradeoff. In either case, I want to work on that feature next, so we should have an answer soon.

> Is there a problem with manually loading runtimes for each model? Shouldn't just making a `loadRuntime(<models_runtime_path>)` call allow it?

Thanks for this question. I'm outlining the approach that I'd put in a design below. I would get feedback on that design separately, as I need to think through some things as outlined in [this other PR comment](https://github.com/triton-inference-server/tensorrt_backend/pull/74#discussion_r1306174948).

We'd need to track the runtimes in a data structure. It's reasonably straightforward in my mind. We would add that data structure. Then, when a model config has a runtime path field, use it to get the runtime path. If that runtime is not already loaded, store it. Then, use that runtime for deserializing any models pointing to the same path. I wanted to make it a separate PR because it is discrete from what this one does, plus it modifies a bit of the managing of runtimes (going from one runtime to multiple).

src/loader.cc
tanmayv25: No issues. I think it is ok to log the setting where it is actually being configured. Offers better debug capability IMO.

src/model_state.h
tanmayv25: I see. Let's keep the version compatibility per-server basis setting.

tanmayv25:
> I think it'd make more sense to generate new models and test version compatibility as a unit test. But perhaps I'm missing something that will change that tradeoff. In either case, I want to work on that feature next, so we should have an answer soon.

I think we should atleast have new models with/without runtime to cover these test cases in CI... We can think about moving all our CI tests to TensorRT version compatibility later.

README.md
```diff
@@ -72,6 +69,13 @@ for detail). Currently the following values are accepted:
   * `BLOCKING`: corresponds to `TRITONBACKEND_EXECUTION_BLOCKING`, this option
   can be set to overlap the host thread workload between model instances.
 
+* The `plugins` flag provides a way to load any custom TensorRT plugins that your models rely on. If you have
+multiple plugins to load, use a semicolon as the delimiter.
+
+* The `version-compatible` flag enables the loading of version-compatible TensorRT models. You must trust the
```
tanmayv25, line 75: Can you link the TensorRT version compatibility feature for users to follow?

tanmayv25 (changes requested):

README.md
dyastremsky: Sure, thanks for suggesting this. Done.

dyastremsky:
> > I think it'd make more sense to generate new models and test version compatibility as a unit test. But perhaps I'm missing something that will change that tradeoff. In either case, I want to work on that feature next, so we should have an answer soon.
> 
> I think we should atleast have new models with/without runtime to cover these test cases in CI... We can think about moving all our CI tests to TensorRT version compatibility later.

Good point, Tanmay. Testing added here: https://github.com/triton-inference-server/server/pull/6255

README.md
```diff
@@ -55,15 +55,12 @@ options.
 
 ##### --backend-config=tensorrt,coalesce-request-input=\<boolean\>,plugins="/path/plugin1.so;/path2/plugin2.so"
```
rmccorm4, line 56: Add `version-compatible` flag to CLI example

dyastremsky: Thanks! Added.

rmccorm4 (approved):

krishung5 (approved):

tanmayv25 (approved):

src/model_state.h
oandreeva-nv, line 142: I think this broke or Windows pipeline: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/68880138

PR #75, opened 2023-09-08
Emable std17 on WIN32

mc-nv:


CMakeLists.txt
```diff
@@ -79,7 +79,11 @@ FetchContent_MakeAvailable(repo-common repo-core repo-backend)
 if(${TRITON_ENABLE_GPU})
   find_package(CUDAToolkit REQUIRED)
   message(STATUS "Using CUDA ${CUDA_VERSION}")
-  set(CUDA_NVCC_FLAGS -std=c++11)
+  if (WIN32)
+    set(CUDA_NVCC_FLAGS -std=c++11)
```
nv-kmcgill53, line 83: At the risk of burying more work in the backlog, can you create a ticket to investigate using C++17 on windows?

nv-kmcgill53 (approved):

CMakeLists.txt
dyastremsky: I don't see these changes already on the main branch. This seems like a recipe for unexpected behavior. Why are we using two different versions of C++ across platforms?

mc-nv: I believe clear answer to your question can be found as a remark on the following page, which is corresponding to the Generator change in https://github.com/triton-inference-server/onnxruntime_backend/pull/212/files. 

Link:
https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170

dyastremsky (approved):

CMakeLists.txt
dyastremsky: Thanks for clarifying. Approving, but let's leave this conversation in case we need to reference when looking at this PR in the future.

PR #80, opened 2024-02-08
Fix dynamic shape outputs

Tabrizian:
When copy output stream is turn on, incorrect allocator for dynamic shape output is used since the allocator is not updated when the binding buffers are being prepared for execution.


rmccorm4 (approved):
Can we have a ticket to add a test that would fail without this fix?

GuanLuo (approved):

rmccorm4:
> Can we have a ticket to add a test that would fail without this fix?

For future reference, this test might have been related? Didn't look too closely: https://github.com/triton-inference-server/server/pull/6057

PR #83, opened 2024-04-30
Remove deprecated APIs in TensorRT backend

pskiran1:
- Deprecated binding index-based APIs were replaced with Tensor name-based
- Removed support for Implicit batch models
- Added support for INT64 data type
- Shape output data type was updated to INT64
- Updated copyright

### DIFF
--- a/src/instance_state.cc
+++ b/src/instance_state.cc
@@ -989,12 +989,26 @@ ModelInstanceState::Run(
       // Custom handling for shape tensors
       // Obtain the shape value
       if (dims.nbDims != 0) {
-        auto shape_value_ptr = reinterpret_cast<int32_t const*>(
+        TRITONSERVER_DataType dt =
+          ConvertTrtTypeToDataType(engine_->getTensorDataType(name.c_str()));
+        if (dt != TRITONSERVER_TYPE_INT64) {
+          FAIL_ALL_AND_RETURN_IF_ERROR(
+              payload_->requests_, payload_->request_count_,
+              payload_->responses_,
+              TRITONSERVER_ErrorNew(
+                  TRITONSERVER_ERROR_INTERNAL,
+                  (std::string("Expected data type of output shape tensor to be INT64 "
+                               "for binding '") +
+                   Name() + "', got " + TRITONSERVER_DataTypeString(dt))
+                      .c_str()),
+              "failed to run TRT response");
+        }
+        auto shape_value_ptr = reinterpret_cast<int64_t const*>(
             citr->second.context_->getTensorAddress(name.c_str()));
 
         // The first shape value must be equal to the total batch_size
         if (support_batching_ &&
-            payload_->total_batch_size_ != (uint32_t)*shape_value_ptr) {
+            payload_->total_batch_size_ != (uint64_t)*shape_value_ptr) {
           FAIL_ALL_AND_RETURN_IF_ERROR(
               payload_->requests_, payload_->request_count_,
               payload_->responses_,
@@ -1122,13 +1136,13 @@ ModelInstanceState::Run(
 
 bool
 ModelInstanceState::SetOutputShapeTensorBuffer(
-    const int32_t* content, TRITONBACKEND_Response** response,
+    const int64_t* content, TRITONBACKEND_Response** response,
     TRITONBACKEND_Output* response_output, const size_t tensor_element_count,
     const int64_t batch_size, cudaStream_t stream)
 {
   bool cuda_copy = false;
 
-  const size_t expected_byte_size = tensor_element_count * sizeof(int32_t);
+  const size_t expected_byte_size = tensor_element_count * sizeof(int64_t);
 
   // Allocate a buffer large enough to hold the serialized tensor.
   TRITONSERVER_MemoryType actual_memory_type = TRITONSERVER_MEMORY_CPU;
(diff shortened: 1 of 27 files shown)


src/model_state.cc
```diff
@@ -504,17 +504,10 @@ ModelState::AutoCompleteConfigHelper(const std::string& model_path)
   }
 
   int max_batch_size = 0;
-  bool has_implicit_batch_dim = false;
-  if (engine->hasImplicitBatchDimension()) {
-    // If engine has implicit batch dimension then retrieve the value and exit
```
Tabrizian, line 509: Could you summarize how implicit batch works?

GuanLuo:
I don't think you need to update the copyright for files that are not under change. Can you remove copyright-only changes?

src/instance_state.cc
```diff
@@ -569,32 +565,30 @@ ModelInstanceState::Run(
...
       continue;
     }
 
-    const std::string& name = engine_->getBindingName(io_index);
-
     TRITONSERVER_Error* err = nullptr;
     // Set the shape binding if needed. If unable to set the shape
     // binding then fail all requests.
-    if (engine_->isShapeBinding(binding_index)) {
+    if (engine_->isShapeInferenceIO(name.c_str())) {
```
GuanLuo, line 585: Is `isShapeInferenceIO` the replacement for `isShapeBinding`? The old code suggests that they are slightly different? CC @tanmayv25

src/instance_state.h
```diff
@@ -410,13 +384,13 @@ class ModelInstanceState : public TensorRTModelInstance {
   // Whether inexact match is allowed for finding CUDA graph
   bool allow_inexact_match_{false};
 
-  // The total number of bindings
-  int total_bindings_{0};
+  // The number of input and output tensors to the model. In case of dynamic
+  // shapes, it is the number of expected input and output tensors to the
+  // configured optimization profile.
+  int total_io_tensors_{0};
```
GuanLuo, line 390: Can you elaborate on number of IO tensor in relation to number of optimization profile? Say if the model has 3 profile and expecting 2 inputs and 2 outputs, is `total_io_tensors_ == 12`?

src/model_state.cc
```diff
@@ -889,15 +872,7 @@ ModelState::FixIO(
 
           // Check if the IO is a shape tensor.
           bool is_shape_tensor = false;
-          int io_index = engine->getBindingIndex(io_name.c_str());
-          if (io_index == -1) {
-            return TRITONSERVER_ErrorNew(
```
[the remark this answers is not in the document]
pskiran1: I think it's safe to remove. We have covered identifying incorrect io names using `CheckAllowedModelInput()` and `CheckAllowedModelOutput()` functions. I have verified the same with sample model configs

https://github.com/triton-inference-server/tensorrt_backend/blob/spolisetty_DLIS-6351/src/model_state.cc#L490-L496
https://github.com/triton-inference-server/tensorrt_backend/blob/spolisetty_DLIS-6351/src/instance_state.cc#L1808-L1821

src/instance_state.cc
pskiran1: Yes, Guan. Based on the TRT documentation, `isShapeInferenceIO` is the replacement for `isShapeBinding`.
https://docs.nvidia.com/deeplearning/tensorrt/migration-guide/index.html#removed-c-api
https://docs.nvidia.com/deeplearning/tensorrt/archives/tensorrt-861/api/c_api/classnvinfer1_1_1_i_cuda_engine.html#a8f257ac5d7d6d7cab2523d81db076b7a

src/instance_state.h
pskiran1: It is the number of input and output tensors for the network from which the engine was built irrespective of the Optimization profiles. It will be initialized by the [getNbIOTensors()](https://docs.nvidia.com/deeplearning/tensorrt/api/c_api/classnvinfer1_1_1_i_cuda_engine.html#af2018924cbea2fa84808040e60c58405). In the above example `total_io_tensors_` will be `4`.

I think I need to update the above comments, they are confusing.

src/model_state.cc
```diff
@@ -663,28 +654,24 @@ ModelState::GetProfileMaxBatchSize(
...
   // minimum batch size supported.
-  for (int binding_index = 0; binding_index < num_profile_bindings;
-       binding_index++) {
-    int effective_binding_index =
-        (profile_index * num_profile_bindings) + binding_index;
-    if (engine->bindingIsInput(effective_binding_index)) {
-      if (!engine->isShapeBinding(effective_binding_index)) {
-        nvinfer1::Dims max_shape = engine->getProfileDimensions(
-            effective_binding_index, profile_index,
+  for (int io_index = 0; io_index < num_io_tensors; io_index++) {
```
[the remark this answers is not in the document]
pskiran1: I believe there is no API to get all tensor names in a single call.
We can only get the names of the IO tensors by calling `getIOTensorName(i) for i in 0 to num_io_tensors`.

Would you recommend storing tensor names in a vector and retrieving them by index instead of calling getIOTensorName(i) multiple times?

src/model_state.cc
pskiran1: Sorry for the late reply. Please refer to https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#explicit-implicit-batch.

Tabrizian (dismissed):

src/instance_state.h
pskiran1: Updated comments.

src/model_state.cc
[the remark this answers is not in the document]
pskiran1: In the `model_state.cc` file, we only use `getIOTensorName(i)` in two places in a loop, and we are not keeping track of engine attributes data related variables. Therefore, have not made any modifications to the `model_state.cc` file.
In the `instance.cc` file, on the other hand, we use `getIOTensorName(i)` in several loops, so added a new variable to store the names and retrieve them.

tanmayv25:
Closing this PR as the TRT 10 related changes will directly go to 24.05 release branches.

PR #91, opened 2024-06-06
Add INT64 Datatype Support for Shape Tensors in TensorRT Backend

pskiran1:
- Introduced `ShapeTensor` class to handle both `int32_t` and `int64_t` shape tensors, using an internal `char` datatype buffer to store the shape tensor values.


src/tensorrt_utils.h
```diff
@@ -31,6 +31,7 @@
 #include <string>
 #include <vector>
 
+#include "instance_state.h"
```
tanmayv25, line 34: I don't think introducing this dependency in a utils header is appropriate. 
I would suggest creating a separate shape_tensor.cc/.h files encapsulating the shape tensor.

src/instance_state.cc
```diff
@@ -1325,12 +1324,6 @@ ModelInstanceState::GetRequestShapeValues(
 
     int io_index = io_index_map_[input_name];
     if (engine_->isShapeInferenceIO(input_name)) {
-      auto it =
```
tanmayv25, line 1328: Why remove this? How are we handling dynamic batching in the shape tensors now?

src/instance_state.cc
```diff
@@ -1359,38 +1352,39 @@ ModelInstanceState::GetRequestShapeValues(
...
-              TRITONSERVER_ERROR_INVALID_ARG,
-              (std::string("shape tensor for input '") + input_name +
-               "' expected byte size is " + std::to_string(expected_byte_size) +
-               " [ or " + std::to_string(expected_byte_size + sizeof(int32_t)) +
-               " if input includes batch shape value] " + ", got " +
-               std::to_string(data_byte_size))
-                  .c_str());
-        }
+      size_t datatype_size;
+      if (datatype == TRITONSERVER_DataType::TRITONSERVER_TYPE_INT32) {
```
tanmayv25, line 1361: Why `if... else...` blocks? Why not a direct `datatype_size=TRITONSERVER_DataTypeByteSize(datatype) `?

src/instance_state.cc
```diff
@@ -1359,38 +1352,39 @@ ModelInstanceState::GetRequestShapeValues(
...
+        datatype_size = TRITONSERVER_DataTypeByteSize(TRITONSERVER_TYPE_INT32);
+      } else if (datatype == TRITONSERVER_DataType::TRITONSERVER_TYPE_INT64) {
+        datatype_size = TRITONSERVER_DataTypeByteSize(TRITONSERVER_TYPE_INT64);
+      } else {
+        return TRITONSERVER_ErrorNew(
+            TRITONSERVER_ERROR_INVALID_ARG,
+            "Un supported shape tensor data type");
```
tanmayv25, line 1368: Unsupported

src/instance_state.cc
```diff
@@ -1549,13 +1542,29 @@ ModelInstanceState::EvaluateTensorRTContext(
               std::abs(opt_dims.d[idx] - input_shape_vec[idx - 1]);
         }
         if (engine_->isShapeInferenceIO(input_name)) {
-          const auto* opt_shape_values = citr->second.opt_shapes_[io_index];
-          *error_distance +=
-              std::abs(*opt_shape_values - (int64_t)total_batch_size);
           auto it = request_shape_values.find(io_index);
-          for (size_t idx = 1; idx < citr->second.nb_shape_values_; idx++) {
+          if (it->second.GetDataType() == ShapeTensorDataType::INT32) {
```
tanmayv25, line 1546: I think the duplicate code can be reduced here. How about adding a function in the ShapeTensor class called `GetDistance` which returns the distance between this shape tensor and the one that gets passed in the argument.
It can take care of data type handling.

src/instance_state.h
```diff
@@ -81,6 +81,50 @@ struct BackendConfiguration {
   bool enable_memory_tracker_{false};
 };
 
+enum class ShapeTensorDataType { INT32, INT64 };
```
tanmayv25, line 84: Moving this to a new leaf node in the design will be cleaner.

src/instance_state.cc
```diff
@@ -1359,38 +1352,39 @@ ModelInstanceState::GetRequestShapeValues(
...
+            TRITONSERVER_ERROR_INVALID_ARG,
+            "Un supported shape tensor data type");
       }
+      const size_t expected_byte_size = element_cnt * datatype_size;
 
-      const int32_t* dims = reinterpret_cast<const int32_t*>(data_buffer);
-      int64_t offset = includes_batch_shape_value ? 1 : 0;
-      for (int64_t i = offset; i < element_cnt; ++i) {
-        it->second.push_back(dims[i]);
+      if ((expected_byte_size != data_byte_size) &&
```
tanmayv25, line 1372: Can we move this check within ShapeTensor class?

src/instance_state.cc
pskiran1: We are handling this as part of `ShapeTensor` while initializing. Please refer https://github.com/triton-inference-server/tensorrt_backend/pull/91/files#diff-8b703b06569a9f63a78b1fbd1985076749c6b1d2bb78b59607a90fa408fd2602R53-R79

src/instance_state.cc
pskiran1: Moved this check to `ShapeTensor::ValidateDataByteSize()` method.

src/instance_state.cc
pskiran1: Moved this code to `ShapeTensor::GetDistance()` method.

src/instance_state.h
pskiran1: Moved `ShapeTensorDataType` to `shape_tensor.h` file. I hope it looks better now. Could you please let me know if any further changes required. Thank you.

tanmayv25 (approved):

PR #96, opened 2024-07-02
Fix tensorrt backend igpu build

pskiran1:
`shape_tensor.cc` and `shape_tensor.h` need to be included only for TRT v10 and above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,17 +122,6 @@ message(STATUS "TRT_VERSION envvar is $ENV{TRT_VERSION}")
 #
 if("$ENV{TRT_VERSION}" VERSION_LESS 10)
   set(SOURCE_DIR "src_trt8")
-else()
-  set(SOURCE_DIR "src")
-endif()
-
-
-#
-# Shared library implementing the Triton Backend API
-#
-configure_file(${SOURCE_DIR}/libtriton_tensorrt.ldscript libtriton_tensorrt.ldscript COPYONLY)
-
-if("$ENV{TRT_VERSION}" VERSION_LESS 10)
   add_library(
     triton-tensorrt-backend SHARED
     ${SOURCE_DIR}/tensorrt.cc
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -132,33 +132,61 @@ endif()
...
-  ${SOURCE_DIR}/loader.cc
-  ${SOURCE_DIR}/loader.h
-  ${SOURCE_DIR}/logging.cc
-  ${SOURCE_DIR}/logging.h
-  ${SOURCE_DIR}/output_allocator.cc
-  ${SOURCE_DIR}/output_allocator.h
-  ${SOURCE_DIR}/io_binding_info.cc
-  ${SOURCE_DIR}/io_binding_info.h
-)
+if("$ENV{TRT_VERSION}" VERSION_LESS 10)
```
fpetrini15, line 135: I would suggest merging this check with the one above. I.e., this:
```
if("$ENV{TRT_VERSION}" VERSION_LESS 10)
  set(SOURCE_DIR "src_trt8")
else()
  set(SOURCE_DIR "src")
endif()
configure_file(${SOURCE_DIR}/libtriton_tensorrt.ldscript libtriton_tensorrt.ldscript COPYONLY)
```
can be merged into the new check your added.

CMakeLists.txt
```diff
@@ -188,6 +178,11 @@ else()
   )
 endif()
 
+#
+# Shared library implementing the Triton Backend API
+#
+configure_file(${SOURCE_DIR}/libtriton_tensorrt.ldscript libtriton_tensorrt.ldscript COPYONLY)
```
fpetrini15, line 184: [optional] It seems like cmake has different scoping rules than C++. In my testing `SOURCE_DIR` will be correctly set in both the if and else execution paths, however, if may be better to place this in the if-else above for clarify.

fpetrini15:
Approved pending successful pipeline. Good work Sai!

fpetrini15 (approved):
Approved pending successful pipeline. Good work Sai!

PR #99, opened 2024-09-04
RHEL SBSA Build

fpetrini15:
**Goal**: Incorporate SBSA builds/tests and extended the smoke tests for x86 and SBSA to include L0_sequence_batcher.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,18 @@ endif()
 set(TRITON_TENSORRT_BACKEND_LIBNAME triton_tensorrt)
 set(TRITON_TENSORRT_BACKEND_INSTALLDIR ${CMAKE_INSTALL_PREFIX}/backends/tensorrt)
 
+set(RHEL_NVINFER_PATH "")
+if(LINUX)
+  file(STRINGS "/etc/os-release" DISTRO_ID_LIKE REGEX "ID_LIKE")
+  if(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
+    if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+      set(RHEL_NVINFER_PATH "/usr/local/cuda/targets/sbsa-linux/lib")
+    else()
+      set(RHEL_NVINFER_PATH "/usr/local/cuda/targets/x86_64-linux/lib")
+    endif()
+  endif(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
+endif(LINUX)
+
 #
 # Dependencies
 #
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -52,6 +52,18 @@ endif()
...
+  file(STRINGS "/etc/os-release" DISTRO_ID_LIKE REGEX "ID_LIKE")
+  if(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
+    if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+      set(RHEL_NVINFER_PATH "/usr/local/cuda/targets/sbsa-linux/lib")
+    else()
+      set(RHEL_NVINFER_PATH "/usr/local/cuda/targets/x86_64-linux/lib")
+    endif()
+  endif(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
+endif(LINUX)
+
```
mc-nv, line 66: can't we extend CMAKE_PREFIX_PATH with predefined locations, to find necessary libraries?

fpetrini15: I suppose we could and it would make things less verbose.

mc-nv (approved):

CMakeLists.txt
```diff
@@ -52,6 +52,8 @@ endif()
 set(TRITON_TENSORRT_BACKEND_LIBNAME triton_tensorrt)
 set(TRITON_TENSORRT_BACKEND_INSTALLDIR ${CMAKE_INSTALL_PREFIX}/backends/tensorrt)
 
+set(CMAKE_PREFIX_PATH "/usr/local/cuda/targets/sbsa-linux/lib;/usr/local/cuda/targets/x86_64-linux/lib")
```
nv-kmcgill53, line 55: 1. Seems like an anti-pattern to have multiple arches in a PATH variable but this is also concise. Maybe add a comment making note that if any other paths need to be added, this should be split. 

2. Does this overwrite any `CMAKE_PREFIX_PATH`s which are added on the command line?

fpetrini15: 1. Circling back on this, I don't like the approach either. I'll set it conditionally.
2. Good catch!

mc-nv: >[Semicolon-separated list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) of directories specifying installation prefixes to be searched by the [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package), [find_program()](https://cmake.org/cmake/help/latest/command/find_program.html#command:find_program), [find_library()](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library), [find_file()](https://cmake.org/cmake/help/latest/command/find_file.html#command:find_file), and [find_path()](https://cmake.org/cmake/help/latest/command/find_path.html#command:find_path) commands. Each command will add appropriate subdirectories (like bin, lib, or include) as specified in its own documentation.

1. In other words if you don't mess with providers, and not planing to provide different architecture CMake will not mess up with dependencies. <br>
There why suppose condition will be redundant  by serving same purpose. 

2. Given argument from command line should extend not override configuration within CMAKE.

fpetrini15: Are you suggesting that we disallow users from setting this variable at all by overwriting it--how it was originally?

e.g., aarch64: `set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} "/usr/local/cuda/targets/sbsa-linux/lib")`

I can't say that if feels right to do that...

nv-kmcgill53 (approved):

PR #100, opened 2024-09-05
Enable Cuda in Graphics Implementation for TensorRT backend

ashishk98:
Add cuda context sharing support for TensorRT backend to reduce context switching overhead when graphics workload is running in parallel


ashishk98:
@nv-kmcgill53 Please review this as discusses

CMakeLists.txt
```diff
@@ -269,6 +269,7 @@ target_link_libraries(
     triton-tensorrt-backend
     PRIVATE
       CUDA::cudart
+      CUDA::cuda_driver
```
nnshah1, line 272: @nv-kmcgill53 , @mc-nv, @tanmayv25  - any issues with this dependency

nnshah1:
@ashishk98 - can you install and run pre-commit checks locally?

```
cd repo; pre-commit install; pre-commit run --all
```

CMakeLists.txt
mc-nv: What is the context behind adding this dependency? 
From documentation is see:
>[CUDA Driver Library](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#cuda-driver-library)
>The CUDA Driver library (cuda) are used by applications that use calls such as `cuMemAlloc`, and `cuMemFree`.
>Targets Created:
>CUDA::cuda_driver
Aren't this dependency is requisite of TensorRT itself? 
Thought by default our product expect driver to be installed and if GPU capability given then available for usage including driver targets and binaries.

nv-kmcgill53: Adding this dependency should be fine. Ashish is linking correctly according to the [cuda documentation](https://docs.nvidia.com/cuda/cuda-runtime-api/driver-vs-runtime-api.html). As it states 
> Context management can be done through the driver API, but is not exposed in the runtime API

So they will need to link against the driver instead of just linking against the cuda runtime.

mc-nv: > So they will need to link against the driver instead of just linking against the cuda runtime.

I'm not agree with this statement, current linkage doesn't explain why user want to add it explicitly.

nv-kmcgill53: The cmake documentation isn't exhaustive when it mentions `cuMemAlloc` and 'cuMemFree`. The user in this case is using the Driver API to set/pass the cuda context around in the backend, rather than letting the `core` take care of this. This is the reason for adding the `CUDA::cuda_driver` lib to the  linking path. This PR necessarily makes use of functions in the driver where the trt_backend didn't before.

mc-nv: Triton TensorRT Backend is unable to work without CUDA, Triton Inference Server and TensorRT installation. 
Current change, per my understanding, uses only `cudaSetDevice` (CUDA::cudart) and `cudaGetErrorString` (CUDA runtime API) and those dependencies are satisfied. There why I don't see any reason to link against `CUDA::cuda_driver`.

ashishk98: The additional dependency is required for cuCtxPushCurrent() / cuCtxPopCurrent() @ https://github.com/triton-inference-server/tensorrt_backend/pull/100/files#diff-3137e95db7c97b9ddd71fa1b600e3dd646025c5a872ad94d4d09f04313fe3fcdR66

ashishk98: The reason we need this dependency is because we are using a special context call Cuda in Graphics (CiG) context which has to work with the cuda driver dll for its operations.

ashishk98:
@nnshah1  fixed pre-commit

ashishk98:
@nnshah1 @mc-nv I have added a new cmake option TRITON_ENABLE_CIG which only conditionally enables the CiG code path as well as conditionally links the cuda_driver component of cuda

src/tensorrt_model.h
```diff
@@ -53,6 +57,41 @@ class TensorRTModel : public BackendModel {
   bool EagerBatching() { return eager_batching_; }
   bool BusyWaitEvents() { return busy_wait_events_; }
 
+#ifdef TRITON_ENABLE_CIG
+  //! Following functions are related to CiG (Cuda in Graphics) context sharing
+  //! for gaming use case. Creating a shared contexts reduces context switching
+  //! overhead and leads to better performance of model execution along side
+  //! Graphics workload.
+  CUcontext GetCiGContext() { return cig_ctx_; }
```
nnshah1, line 65: @ashishk98 question: is this specific to CIG - or could be applied to any application provided cuda context?

src/tensorrt_model.cc
```diff
@@ -90,6 +96,23 @@ TensorRTModel::ParseModelConfig()
     }
   }
 
+#ifdef TRITON_ENABLE_CIG
+  triton::common::TritonJson::Value parameters;
+  if (model_config_.Find("parameters", &parameters)) {
+    triton::common::TritonJson::Value value;
+    std::string ptr_value;
+    if (parameters.Find("CIG_CONTEXT_PTR", &value)) {
+      RETURN_IF_ERROR(value.MemberAsString("string_value", &ptr_value));
```
nnshah1, line 105: @ashishk98 instead of directly converting here as a special case, I would prefer to use something similar to what is done in the trt-llm backend:

https://github.com/triton-inference-server/tensorrtllm_backend/blob/8ffb174c0fe88e677eeed7928348e20be548f3f6/inflight_batcher_llm/src/model_state.cc#L204 

In this case there is a template method to convert from a parameter to a value - I think the code will be a little clearer to follow.

Also - can we convert to and from a 64 bit integer?

so something like:

```
model_state->GetParameter<uint64>("CUDA_CONTEXT");
```

Also it strikes me that although we use ```value.MemberAsString()```

we could also use ```value.MemberAsUint("string_value",&ptr_value)```

Instead (https://github.com/triton-inference-server/common/blob/578491fc3944f77d16a6a38e3d7691c485c47ba0/include/triton/common/triton_json.h#L927)

So two things - 1) add a templated GetParameter() method and 2) we can use MemberAsUint for the uint64 template. 3) officially transfer uint64 values and convert them to and from context.

src/model_state.cc
```diff
@@ -373,13 +390,19 @@ ModelState::AutoCompleteConfig()
...
-  cuerr = cudaSetDevice(current_device);
-  if (cuerr != cudaSuccess) {
-    return TRITONSERVER_ErrorNew(
-        TRITONSERVER_ERROR_INTERNAL,
-        (std::string("unable to revert CUDA device to GPU ") +
-         std::to_string(current_device) + " : " + cudaGetErrorString(cuerr))
-            .c_str());
+#ifdef TRITON_ENABLE_CIG
+  // Set device if CiG is disabled
+  if (!isCiGEnabled())
```
nnshah1, line 395: @tanmayv25, @ashishk98  - is there a way to have a single scoped object

```ScopedCudaDeviceContext``` 

That internally checks if there is an 'application_context` and if there is an application context uses push / pop - if not uses `cudaSetDevice` ?

We don't currently use them in the same locations - but am wondering if that would be possible - I think it would be cleaner logically - where basically an 'application_context' takes the place of the 'device' but otherwise the logic remains the same.

```
    ScopedObject(Device);
    ScopedObject(Context);
```

File: src/model_state.cc [hunk 5]
```diff
@@ -175,7 +175,13 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
 ModelState::~ModelState()
 {
   for (auto& device_engine : device_engines_) {
-    cudaSetDevice(device_engine.first.first);
+#ifdef TRITON_ENABLE_CIG
```
nnshah1, line 178: I know I had asked for looking at macros to enable, but I would like to avoid this kind of guard - if we can use a single method and then have two different implementations of that method / object would prefer that to having the macros embedded in the functions / methods.

nnshah1:
@tanmayv25 , @ashishk98 

I would like to see:

1) Can we generalize this to: "TRITON_ENABLE_APPLICATION_CONTEXT" instead of CIG - as it looks to be general for any cuda context?

2) Can we use a more generic `GetParameter<uint64>` templated method similar to tensorrtllm backend. For the future we may consider adding this to the backend apis as it seems like a common pattern. small optimization here could be to use MemberAsUint() instead of MemberAsString in the <uint64> template

3) Can we create a method / function that would selectively use setdevice or push/pop but otherwise be the same logic? We can then enable the alternate constructor / check in one place instead of multiple.

src/instance_state.cc
```diff
@@ -424,7 +430,13 @@ ModelInstanceState::Run(
   payload_.reset(new Payload(next_set_, requests, request_count));
   SET_TIMESTAMP(payload_->compute_start_ns_);
 
-  cudaSetDevice(DeviceId());
+#ifdef TRITON_ENABLE_CIG
+  // Set device if CiG is disabled
+  if (!model_state_->isCiGEnabled())
+#endif  // TRITON_ENABLE_CIG
+  {
+    cudaSetDevice(DeviceId());
```
GuanLuo, line 438: Do you mind to share the reasoning of avoiding the set device calls? Wouldn't that cause the issue of model not being placed / executed on selected device (based on model config)?

tanmayv25:
How model instances on multiple GPUs will be handled? AFAIK a cuda context is per device. If we have more than one GPU devices, then we should pass cuda context handles for each model instance. Or am I missing something here?

CMakeLists.txt
```diff
@@ -271,6 +273,17 @@ target_link_libraries(
...
+if(${TRITON_ENABLE_CIG})
+  target_compile_definitions(
+      triton-tensorrt-backend
+      PRIVATE TRITON_ENABLE_CIG
+  )
+  target_link_libraries(
+      triton-tensorrt-backend
+      PRIVATE
+        CUDA::cuda_driver
+  )
```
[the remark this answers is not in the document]
ashishk98: What is a generator expression?

src/instance_state.cc
ashishk98: 1. The intended use of cuda context sharing is targeted only of single GPU (RTX end-user) systems. I wanted to avoid complications with this use case
2. When we call cudaSetDevice() the cuda runtime resets the to using the default cuda context for the thread

src/model_state.cc
ashishk98: Fixed

src/model_state.cc
ashishk98: We can take a look at this in the next iteration

src/tensorrt_model.cc
ashishk98: I have added a GetParameter call for std::string instead of UINT64. This is because when we add the parameter to model config it is directly converted into a hex string instead of a numeric string. Hence while parsing the pointer, MemberAsUint fails because it gets a hex string to parse.

tanmayv25:
@ashishk98 I believe we still need to raise an error if someone tries to use pre-built cuda context with multi-GPU environment, right?

src/tensorrt_model.cc
```diff
@@ -120,4 +128,28 @@ TensorRTModel::GetCudaStreamPriority()
   return cuda_stream_priority;
 }
 
+template <>
+TRITONSERVER_Error*
+TensorRTModel::GetParameter<std::string>(
+    std::string const& name, std::string& str_value)
+{
+  triton::common::TritonJson::Value parameters;
+  TRITONSERVER_Error* err =
```
tanmayv25, line 137: Use RETURN_IF_ERROR Macro.

src/tensorrt_model.cc
```diff
@@ -120,4 +128,28 @@ TensorRTModel::GetCudaStreamPriority()
...
+      model_config_.MemberAsObject("parameters", &parameters);
+  if (err != nullptr) {
+    return err;
+    // throw std::runtime_error("Model config doesn't have a parameters
+    // section");
+  }
+  triton::common::TritonJson::Value value;
+  err = parameters.MemberAsObject(name.c_str(), &value);
```
tanmayv25, line 145: Use RETURN_IF_ERROR macro

src/tensorrt_model.h
```diff
@@ -53,6 +66,65 @@ class TensorRTModel : public BackendModel {
...
+  {
+    std::stringstream ss;
+    ss << str;
+
+    void* ctx_ptr;
+    ss >> ctx_ptr;
+    return ctx_ptr;
+  }
+
+  //! Following functions are related to Cuda (Cuda in Graphics) context sharing
```
tanmayv25, line 83: related to custom cuda context sharing?

fpetrini15:
Can the stakeholders provide another round of reviews on this PR? We'd like to get these changes into a release asset this week.

tanmayv25:
> @ashishk98 I believe we still need to raise an error if someone tries to use pre-built cuda context with multi-GPU environment, right?

I still don't see a guard against preventing the use of pre-built cuda context on a multi-GPU system. We are skipping calling cudaSetDevice() when using CiG context. And as we are passing only a single CiG context all the model instances will hit the same GPU device. We need to put a check if Triton is loading model instance on different device and throw a meaningful error.

ashishk98:
@tanmayv25 please review the latest commit

src/tensorrt_model.h
```diff
@@ -61,6 +133,30 @@ class TensorRTModel : public BackendModel {
...
+  CUcontext cuda_ctx = nullptr;
+#endif  // TRITON_ENABLE_CUDA_CTX_SHARING
+};
+
+struct ScopedRuntimeCudaContext {
+  ScopedRuntimeCudaContext(TensorRTModel* model_state)
+      : model_state_(model_state)
+  {
+#ifdef TRITON_ENABLE_CUDA_CTX_SHARING
+    if (model_state_->isCudaContextSharingEnabled()) {
```
tanmayv25, line 146:
```suggestion
    if (model_state_->IsCudaContextSharingEnabled()) {
```

src/tensorrt_model.h
```diff
@@ -61,6 +133,30 @@ class TensorRTModel : public BackendModel {
...
+      THROW_IF_BACKEND_MODEL_ERROR(model_state_->PushCudaContext());
+    }
+#endif  // TRITON_ENABLE_CUDA_CTX_SHARING
+  }
+  ~ScopedRuntimeCudaContext()
+  {
+#ifdef TRITON_ENABLE_CUDA_CTX_SHARING
+    if (model_state_->isCudaContextSharingEnabled()) {
```
tanmayv25, line 154:
```suggestion
    if (model_state_->IsCudaContextSharingEnabled()) {
```

tanmayv25:
Some minor fixes, otherwise LGTM.

ashishk98:
Resolved minor fixes as well

tanmayv25 (approved):

tanmayv25:
@nvda-mesharma did you verify these changes in CI? The pipeline I launched yesterday seems to be failing with this error?

```
gmake[2]: *** [CMakeFiles/triton-tensorrt-backend.dir/build.make:202: CMakeFiles/triton-tensorrt-backend.dir/src/loader.cc.o] Error 1
In file included from /tmp/tritonbuild/tensorrt/src/model_state.h:33,
                 from /tmp/tritonbuild/tensorrt/src/instance_state.h:36,
                 from /tmp/tritonbuild/tensorrt/src/instance_state.cc:27:
/tmp/tritonbuild/tensorrt/src/tensorrt_model.h:69:13: error: explicit specialization in non-namespace scope 'class triton::backend::tensorrt::TensorRTModel'
   69 |   template <>
      |             ^
/tmp/tritonbuild/tensorrt/src/tensorrt_model.h:70:23: error: template-id 'GetParameter<std::string>' in declaration of primary template
   70 |   TRITONSERVER_Error* GetParameter<std::string>(
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /tmp/tritonbuild/tensorrt/src/model_state.h:33,
                 from /tmp/tritonbuild/tensorrt/src/instance_state.h:36,
                 from /tmp/tritonbuild/tensorrt/src/tensorrt.cc:39:
/tmp/tritonbuild/tensorrt/src/tensorrt_model.h:69:13: error: explicit specialization in non-namespace scope 'class triton::backend::tensorrt::TensorRTModel'
   69 |   template <>
      |             ^
/tmp/tritonbuild/tensorrt/src/tensorrt_model.h:70:23: error: template-id 'GetParameter<std::string>' in declaration of primary template
   70 |   TRITONSERVER_Error* GetParameter<std::string>(
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/tritonbuild/tensorrt/src/instance_state.cc: In member function 'void triton::backend::tensorrt::ModelInstanceState::ProcessResponse()':
/tmp/tritonbuild/tensorrt/src/instance_state.cc:1216:29: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
 1216 |     auto payload = std::move(completion_queue_.Get());
```

https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/137650575

@ashishk98  for viz.

PR #106, opened 2025-02-03
TensorRT Cuda Context Sharing

ashishk98:
This PR enabled cuda context sharing with CiG streams on Windows Gaming applications


CMakeLists.txt
mc-nv, line 1: This change is Windows only as per my understanding, `CMakeLists.txt` doesn't reflect it and it cause a confusion. 
Is it something that we consider to build by default then configuration change should reflect it accordingly.

src/instance_state.cc
mc-nv, line 1: Files in `src` folder may be related not only to Windows but other platforms as well. We need to make sure it doesn't impact the container build.

@GuanLuo are we OK with it? <br>
@saturley-hall have we tested it well in CI for other stacks are we confident with result?

PR #108, opened 2025-04-02
feat: Support for allocating GPU memory based on the selected profile

anmolgupt:
The changes in the PR support 2 main items:

1. the GPU memory is allocated based on the selected TensorRT profile and not based on the profile that consumes max memory even when it's not selected.
2. Avoid the creation of profile 0 execution context if it's required.

### DIFF
--- a/src/instance_state.cc
+++ b/src/instance_state.cc
@@ -1,4 +1,4 @@
-// Copyright 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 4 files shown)


dongfengy:
For kUSER_MANAGED , the user (in this case the triton server) would need to actually allocate a piece of device memory and pass to execution context. Do you have support for this behavior? If not I would suggest you to only add kSTATIC and kON_PROFILE_CHANGE

src/model_state.cc
```diff
@@ -288,6 +289,42 @@ ModelState::ValidateModelConfig()
...
 ModelState::ParseParameters()
 {
+  triton::common::TritonJson::Value params;
+  bool status = ModelConfig().Find("parameters", &params);
+  if (status) {
+    // If 'allocation_strategy' is not present in 'parameters',
+    // will use the default strategy "STATIC".
+    std::string alloc_strategy;
+    TRITONSERVER_Error* err =
+        GetParameterValue(params, "allocation_strategy", &alloc_strategy);
```
yinggeh, line 299: @anmolgupt Shortened parameter key from "execution_context_allocation_strategy" to "allocation_strategy". Let me know if it causes any ambiguity.

src/instance_state.cc
```diff
@@ -1697,15 +1697,6 @@ ModelInstanceState::InitOptimizationProfiles()
   // the first context creation. As currently triton supports one
   // context per engine, in order to set the specified profile_index,
   // another context is created and the previous context is destroyed.
```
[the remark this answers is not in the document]
anmolgupt: I tested at my end, the changes work for me.

src/model_state.cc
```diff
@@ -288,6 +289,42 @@ ModelState::ValidateModelConfig()
 TRITONSERVER_Error*
 ModelState::ParseParameters()
 {
+  triton::common::TritonJson::Value params;
+  bool status = ModelConfig().Find("parameters", &params);
+  if (status) {
+    // If 'allocation_strategy' is not present in 'parameters',
```
yinggeh, line 295:
```suggestion
    // If 'execution_context_allocation_strategy' is not present in 'parameters',
```

src/model_state.cc
```diff
@@ -288,6 +289,42 @@ ModelState::ValidateModelConfig()
...
+    TRITONSERVER_Error* err =
+        GetParameterValue(params, "execution_context_allocation_strategy", &alloc_strategy);
+    if (err != nullptr) {
+      if (TRITONSERVER_ErrorCode(err) != TRITONSERVER_ERROR_NOT_FOUND) {
+        return err;
+      } else {
+        TRITONSERVER_ErrorDelete(err);
+      }
+    } else {
+      // allocation_strategy is present in model config parameters
```
yinggeh, line 307:
```suggestion
      // 'execution_context_allocation_strategy' is present in model config parameters.
```

anmolgupt:
@yinggeh: New changes look good to me; I got the expected results on the models with these updates.

yinggeh:
Updated document

yinggeh:
LGTM. Thanks for your contribution.

yinggeh (approved):

PR #110, opened 2025-05-03
fix: Proper handling of shape tensor size

yinggeh:
#### What does the PR do?
Add check to prevent integer underflow of shape tensor size.

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [x] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [ ] Added [test plan](#test-plan) and verified test passes.
- [x] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] fix

#### Test plan:

- CI Pipeline ID:

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- closes GitHub issue: #xxx

### DIFF
--- a/src/shape_tensor.cc
+++ b/src/shape_tensor.cc
@@ -1,4 +1,4 @@
-// Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions


src/shape_tensor.cc
```diff
@@ -72,6 +72,11 @@ ShapeTensor::SetDataFromBuffer(
       *reinterpret_cast<int64_t*>(data_.get()) =
           static_cast<int64_t>(total_batch_size);
     }
+    if (size_ < datatype_size) {
```
tanmayv25, line 75: This would only be possible for empty shape tensors. That is nb_shape_values_ is 0, correct?
Does the logic even come here at this place?
Do you have a test case for this?

tanmayv25 (approved):
Approving as it seems an innocuous check.

src/shape_tensor.cc
yinggeh: I couldn't come up with a test case because I don't think we will have any overflow here. Core will validate if input shape matches model config. And I could not create a plan file with empty dimension input.

PR #118, opened 2026-01-06
fix: Optimize TensorRT backend to avoid duplicate host-memory allocation during engine deserialization

pskiran1:


yinggeh (dismissed):

yinggeh:
Need to update copyrights

yinggeh (approved):

PR #127, opened 2026-09-03
feat: check request cancellation before batch assembly

pskiran1:
## What does the PR do?

The PR checks `TRITONBACKEND_RequestIsCancelled` for each request in `Run()` and, when every request in the batch has been cancelled, responds and returns without submitting any work to the engine.

TensorRT issues a batch as a single asynchronous submission — one `enqueueV3`, a `cudaGraphLaunch` replay, or a multi-device fan-out — so there is no intermediate point at which partial work can be abandoned. The only actionable checkpoint is before the batch is assembled, which is where this check sits.

Scope is deliberately whole-batch only. Dropping a subset of requests from a forming batch would require re-deriving the batch size, shape tensor values and optimization profile selection, and is not worth the complexity unless profiling shows a meaningful rate of mid-batch cancellations.

## Related PRs
- core: https://github.com/triton-inference-server/core/pull/522 — check request cancellation before backend dispatch
- server: https://github.com/triton-inference-server/server/pull/8953 — `L0_request_cancellation` coverage


src/instance_state.cc
```diff
@@ -848,6 +848,8 @@ ModelInstanceState::Run(
   // execution. The batch-size, number of inputs, and size of each
   // input has already been checked so don't need to do that here.
   payload_->total_batch_size_ = 0;
+  bool all_requests_cancelled =
```
yinggeh: Is `all_requests_cancelled` a descriptive name for the value assigned here?

pskiran1: The initial value represents whether execution may be skipped, not whether all requests are already known to be cancelled. I renamed `all_requests_cancelled` to `skip_batch_execution` for better naming convention.

Vinya567 (dismissed):

whoisj (dismissed):

src/instance_state.cc
```diff
@@ -848,6 +848,8 @@ ModelInstanceState::Run(
   // execution. The batch-size, number of inputs, and size of each
   // input has already been checked so don't need to do that here.
   payload_->total_batch_size_ = 0;
+  bool skip_batch_execution =
```
yinggeh:
```suggestion
  bool may_skip_execution =
```

yinggeh: `skip_batch_execution` reads like a decision already made, and “batch” is misleading because non-batch models (`max_batch_size == 0`) take this same path.

The initializer is only eligibility: skip is allowed when the payload is non-empty and skipping will not drop an implicit-state update. After the loop it becomes the actual skip only if every request was cancelled.

`may_skip_execution` covers both meanings — allowed at the start, still possible after each request — without implying we skip just because the model has no implicit state.

yinggeh: Also, move these lines next to where it's first used.

src/instance_state.cc
```diff
@@ -848,6 +848,8 @@ ModelInstanceState::Run(
...
+      (payload_->request_count_ > 0) && !uses_implicit_state_;
```
yinggeh, line 852: `uses_implicit_state_` is set only when `sequence_batching.state` is present. Sequence models that only use explicit control tensors (START/END/READY) can still skip an all-cancelled payload.

src/instance_state.cc
pskiran1: Done. Renamed it to may_skip_execution and clarified the comments.

src/instance_state.cc
pskiran1: Yes, this is intentional. We only prevent skipping when `sequence_batching.state` is configured.

whoisj (approved):
Approved, but give @yinggeh 24 hours to respond to the changes made for him as well as the replies to him.

yinggeh (approved):

PR #129, opened 2026-09-11
build: Update C++ standard from c++17 to c++20 (#128)

mc-nv:
Cherry-pick of the C++17->C++20 standard bump (TRI-1877) onto `r26.09`.

* build: Update C++ standard from c++17 to c++20

Upstream PyTorch raised ATen's minimum required C++ standard from 17
to 20 (pytorch/pytorch#178150), which broke the pytorch_backend build.
Align this repo's TRITON_MIN_CXX_STANDARD/CMAKE_CXX_STANDARD default
with that floor for consistency across the stack.

TRI-1877

* fix: rename semaphore.h to avoid colliding with POSIX <semaphore.h>

Under C++20, libstdc++'s <condition_variable> transitively needs the
POSIX <semaphore.h> to support std::counting_semaphore. This repo's
own src/semaphore.h has the same name and sits ahead of the system
include path (-I<src> precedes default system dirs), so the compiler
resolved the internal #include <semaphore.h> to this project's own
header instead of glibc's, which defines no POSIX semaphore symbols.
That broke the C++17->20 bump build with a cascade of "sem_t does not
name a type" / "condition_variable does not name a type" errors.

Rename to triton_semaphore.h so it can no longer shadow the system
header; harmless under C++17 since nothing pulled in <semaphore.h>
there.

TRI-1877

* fix: rename filesystem.h/.cc to avoid colliding with <filesystem>

Same class of bug as the semaphore.h rename in this repo: a project
header sharing a name with a C++ standard library header, combined
with -I<src> preceding the system include path, means any future
internal #include <filesystem> from libstdc++ (or a toolchain bump)
would silently resolve to this project's own header instead. Not
currently triggering a build failure, but proactively renamed given
this repo already proved the pattern is live once.

TRI-1877

(cherry picked from commit 38b69d018702d482608dae46c9775ab27f2007e1)


src/triton_filesystem.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2022-2026, NVIDIA CORPORATION. All rights reserved.
```
yinggeh:
```suggestion
// Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
```

src/triton_filesystem.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2022-2026, NVIDIA CORPORATION. All rights reserved.
```
yinggeh:
```suggestion
// Copyright (c) 2022, NVIDIA CORPORATION. All rights reserved.
```

src/triton_semaphore.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
yinggeh:
```suggestion
// Copyright 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

yinggeh (changes requested):

yinggeh (approved):

